PR #61, opened 2015-10-19
added option to enable local debug info by adding --compile-zinc-debug-symbol to pants

wisechengyi:


wisechengyi:
depending on https://github.com/pantsbuild/pants/pull/2428

wisechengyi:
adding to pants.ini is more viable

PR #64, opened 2015-10-28
 Show a pop up to install Python Plugin if not installed upon BUILD file inspection

wisechengyi:
remind user to install python plugin for pants, and one click to take them to the plugin window


wisechengyi:
merged as b59e7afb8551b3c1d33b776c2b2791350f031dbc

PR #65, opened 2015-10-29
corrected target name extraction; resolution returns file element if target element is not found

wisechengyi:
sanitized ':' for resolution; disabled highlighting on unresolved target by returning a dummy target


wisechengyi:
merged https://github.com/pantsbuild/intellij-pants-plugin/commit/1175873344b7cfb0141f4ddb5471eafb6f5b0ed0

PR #66, opened 2015-10-30
build file type inspection, and quick fix to treat build files as python files

wisechengyi:
build file type inspection, and quick fix to treat build files as python files


wisechengyi:
merged as ce57efc6bf3290bd53978ec3e77cf866935bba06

PR #67, opened 2015-10-31
send SIGTERM to pants for clean exit instead of force kill upon run cancellation

wisechengyi:
send SIGTERM to pants for clean exit instead of force kill upon run cancellation


wisechengyi:
merged as 4eca8209fcfce03a0e792e77f2cfadbbe28f53a3

PR #68, opened 2015-11-06
Fix for "@NotNull parameter 'defaultValue' of com/intellij/util/ObjectUtils.notNull must not be null"

wisechengyi:
Straight forward check instead of relying on notNull function assumptions.


wisechengyi:
merged as 43e0332b4e110d8cfafdd981b07cff44f28ab31f

PR #69, opened 2015-11-06
readme includes python plugin in installation

wisechengyi:
Straight forward check instead of relying on notNull function assumptions.


wisechengyi:
discarded

PR #72, opened 2015-11-20
fix stale pants flags to load sources and docs for intelij pants plugin

wisechengyi:
Fix the issue where 3rdparty sources cannot be resolved by intellij because the flags are staled.
This change also includes loading docs along with sources once the "load sources and docs for libraries" box is checked.


wisechengyi:
merged as 5c08585

PR #73, opened 2015-11-20
Fix IntelliJ CI test case based on pants dependencies rather than hardcoding module names

wisechengyi:
https://github.com/pantsbuild/pants/commit/a31e175730dad9b6694204b4de7fbb9906542cbb removed a target under examples/src/java/org/pantsbuild/example/jaxb/main, thus 
    
doImport("examples/src/java/org/pantsbuild/example/jaxb/main");

assertModules(
  "examples_src_resources_org_pantsbuild_example_jaxb_jaxb",
  "examples_src_resources_org_pantsbuild_example_names_names", <- this module is gone
  "examples_src_java_org_pantsbuild_example_jaxb_main_main",
  "examples_src_java_org_pantsbuild_example_jaxb_reader_reader"
);

Pinning pants version at 0.0.59 instead of pulling the latest OSS pants for CI.


wisechengyi:
merged as 1b151f1180b29ba482491cdcdd32953f3d9f01bf

PR #76, opened 2015-11-29
remove writeAction thread on folder refresh as there is nothing to write

wisechengyi:
writeAction is only deployed to lock the data to write, and it does not apply in this case.

### DIFF
--- a/src/com/twitter/intellij/pants/execution/PantsClasspathRunConfigurationExtension.java
+++ b/src/com/twitter/intellij/pants/execution/PantsClasspathRunConfigurationExtension.java
@@ -55,18 +55,11 @@ public class PantsClasspathRunConfigurationExtension extends RunConfigurationExt
       classpath.remove(excludedPath);
     }
 
-    ApplicationManager.getApplication().runWriteAction(
-      new Runnable() {
-        @Override
-        public void run() {
-          final VirtualFile workingDir = PantsUtil.findPantsWorkingDir(module);
-          if (workingDir != null) {
-            // we need to refresh because IJ might not pick all newly created symlinks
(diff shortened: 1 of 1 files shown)


fkorotkov:
have you tested it? Im' pretty sure it will throw an assertion.

All invokations are wrapped in write action. See https://upsource.jetbrains.com/idea-ce/file/idea-ce-40e7343ce3dee47a31b53bb7af78532473351965/platform%2Flang-impl%2Fsrc%2Fcom%2Fintellij%2Futil%2Fdownload%2Fimpl%2FFileDownloaderImpl.java?nav=12869:12892&findUsages=%2Fplatform%2Fcore-api%2Fsrc%2Fcom%2Fintellij%2Fopenapi%2Fvfs%2FVirtualFileManager.java:105

wisechengyi:
yes manually. in the link, probably because
"result.setResult(VirtualFileManager.getInstance().refreshAndFindFileByUrl(url));"
is a write action? refresh itself should not write anything.

-Yi

fkorotkov:
Refresh might do some writes for newly founded files. You probably didn't see an assertion because there was no attempt to write. 

See [JavaDoc for the method](https://upsource.jetbrains.com/idea-ce/file/idea-ce-40e7343ce3dee47a31b53bb7af78532473351965/platform%2Fcore-api%2Fsrc%2Fcom%2Fintellij%2Fopenapi%2Fvfs%2FVirtualFileManager.java?nav=4034:4057:focused)(line 103): *This method should be only called within write-action.*

wisechengyi:
Yeah that was very likely what happened, and it's good to go along with the docs. Thanks for the pointers!

PR #77, opened 2015-12-03
Add pants run/debug configuration to method and package level. Tests added as well

wisechengyi:
This change will allow testing on a particular a method by right clicking on the method, so users do not have to run all the tests in a class, and on a package, so they don't have to run each class individually.


wisechengyi:
8d03aff680c4325fb9b26873ca62151279ce3212

PR #80, opened 2015-12-05
temporary way to let users know which version of IJ is compatible the plugin

wisechengyi:
Currently there is no way to let users know which version of IJ is good for the plugin. The long term solution is https://github.com/pantsbuild/pants/issues/2665


wisechengyi:
merged as 58c3a30653337c038c168ff874065e6da0833ed1

wisechengyi:
58c3a30653337c038c168ff874065e6da0833ed1

PR #82, opened 2015-12-11
correct jps recompile logic; catch compile error; temp fix to filter out _synthetic_resources

wisechengyi:


wisechengyi:
merged as 02c92ef980ed4f4ab5c2b9efe386f266fd3299ad

PR #84, opened 2015-12-15
Have regression cover the last 10 releases of pants; use git tag to specify the pants version to load instead of sha 

wisechengyi:
@zundel mentioned to specify a policy regarding how far back the regression test should go, and I think 10 releases make sense which goes back about 2-3 months.


wisechengyi:
merged as 5d81ae19ad800f109346133e4a2c0c014bee7d35

PR #85, opened 2015-12-15
readme update: do not import plugin code with the plugin itself

wisechengyi:
+ Do not import the plugin source as pants project because the plugin does not work on itself.


wisechengyi:
100675d5dad9b0668de3bb7bcd5d4fbaf7294377

PR #86, opened 2015-12-15
Using the correct Virtual File System refresh function before loading classpath

wisechengyi:
According to http://grepcode.com/file/repository.grepcode.com/java/ext/com.jetbrains/intellij-idea/10.5/com/intellij/openapi/vfs/VirtualFileManager.java#VirtualFileManager.refreshWithoutFileWatcher%28boolean%29

Prevoisly used refreshAndFindFileByUrl "refreshes only the part of the file system needed for searching the file by the given URL and finds file by the given URL." This method will only non-recursively refresh, /a, /a/b/, to find /a/b/c. 

The symptom is that, symlink /a/b/c/123 -> .pants.d/123 changes to /a/b/c/456 -> .pants.d/456, VFS still think /a/b/c/123 exists, resulting classpath not found in junit run. Ideally we want to refresh /a/b/c and recursively of its content.

Therefore, using refreshWithoutFileWatcher as it "Refreshes the cached information for all files in this file system from the physical file system" com.intellij.openapi.vfs.VirtualFileSystem#refresh

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,10 +7,10 @@ sudo: false
 language: java
 jdk:
   - oraclejdk7
-#cache:
-#  directories:
-#  - .cache/intellij
-#  - .cache/bootstrap
+cache:
+  directories:
+  - .cache/intellij
+  - .cache/bootstrap
 
 notifications:
   email:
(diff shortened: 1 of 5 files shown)


wisechengyi:
CI this

wisechengyi:
merged as 4cd4e1c2d03f3b5e1820e283ac19bbe21dc62e85

wisechengyi:
4cd4e1c2d03f3b5e1820e283ac19bbe21dc62e85

PR #87, opened 2015-12-18
remove --no-test-junit-suppress-output, leave the choice to user

wisechengyi:
1. "--no-test-junit-suppress-output" is no longer valid as it is replaced by "-test-junit-output-mode"
2. we want to leave the choice to user instead of implicitly injecting this flag. If you want it by default, add the following to pants.ini 
```
[test.junit]
output_mode: FAILURE_ONLY
```


wisechengyi:
92304f560df82922f135a4a476b8abed13c7baf5

PR #88, opened 2015-12-22
reorganize setup script; add hash check for downloaded packages; disable cache

wisechengyi:
Earlier unchecked packages can cause runtime failure with obscure error messages. This change addresses the setup part, so the errors seen later will not be related to setup. 

This change moves all the env var setup to scripts/prepare-ci-environment.sh, and adds md5 checks into scripts/setup-ci-environment.sh. md5sum is used because it is the same command on linux and osx, whereas sha256 is different.

Cache is also disabled as it only provides 10-20 secs speed up(whole ci runs ~15 min), but adds another layer of compilication.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,10 +7,10 @@ sudo: false
 language: java
 jdk:
   - oraclejdk7
-cache:
-  directories:
-  - .cache/intellij
-  - .cache/bootstrap
+#cache:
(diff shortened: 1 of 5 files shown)


wisechengyi:
bb71bb1d5fd97dff74f1642a04570f3d38e8aafc

PR #96, opened 2016-01-08
pick up jars with custom classifiers; test added

wisechengyi:
Earlier if a library is like this:
```
    "org.apache.hadoop:hadoop-common:2.7.1": {
        "default": "/Users/yic/.ivy2/pants/org.apache.hadoop/hadoop-common/jars/hadoop-common-2.7.1.jar",
        "sources": "/Users/yic/.ivy2/pants/org.apache.hadoop/hadoop-common/sources/hadoop-common-2.7.1-sources.jar",
        "tests": "/Users/yic/.ivy2/pants/org.apache.hadoop/hadoop-common/jars/hadoop-common-2.7.1-tests.jar",
        "javadoc": "/Users/yic/.ivy2/pants/org.apache.hadoop/hadoop-common/javadocs/hadoop-common-2.7.1-javadoc.jar"
    },
```
Only "default", "sources", and "javadoc" jars will be imported into the project, whereas the jar with custom classifier "tests" is ignored.

This change will import all jars with custom classifiers as binary jars.

From the code perspective:
1. LibraryInfo used to static fields default, sources, and javadoc, but custom classifiers can be anything and there can be any number of them. Now the class extends HashMap to handle the fields dynamically while maintaining the interface that LibraryInfo used to support.
2. A custom deserializer is introduced for LibraryInfo to parse the export output.


wisechengyi:
4298cf30536b72091af821dfe9e5523c5b1e69b6

PR #97, opened 2016-01-12
Abort importing directory or target if it uses a different pants executable, i.e. in different source root

wisechengyi:
Earlier if user inadvertently references different local repo (from the original project's target) when importing a pants target-based module to a project, IntelliJ sometimes references the wrong repo in the full source paths.

This change will enforce one pants executable per project by checking the pants executable at the time of import.


wisechengyi:
34a648ae9799f0a08fcdfb93e14c91beb0c094ce

PR #101, opened 2016-01-21
UX enhancement: auto zoom in at the target/directory imported

wisechengyi:
Currently when a target is imported, for example testprojects/tests/java/org/pantsbuild/testproject as below, there is no indication where the project files are unless user remembers the filename to search for. 
screen shot 2016-01-21 at 12 09 54 pm

This change will automatically open a file directly under the folder imported, so user can click 'scroll from source' to locate its position in any view preferred as below.

screen shot 2016-01-23 at 1 57 01 am


wisechengyi:
CI this

wisechengyi:
e0fa7b5d223f6502e1239334c809812f8afaabff

PR #102, opened 2016-01-26
Honor JDK selected during import for compilation

wisechengyi:
Currently the JDK selected during import stage is only used as code reference and does not factor into pants compilation which uses the system default JDK version. This change will let the plugin honor the JDK selected by using --jvm-distributions-paths


wisechengyi:
f9ad862291e5b9d99493395380f710fb76eb66c2

PR #104, opened 2016-02-02
Refactor Globs/TargetAddressInfo into common module

wisechengyi:
Earlier symlinks/duplications were created to simplify the logic in https://rbcommons.com/s/twitter/r/3300/ , so this RB corrects the duplication and refactors Globs/TargetAddressInfo into common module


wisechengyi:
abdd988cb6ac8e86dc418d0e060f510c9d503443

PR #106, opened 2016-02-04
Remove PantsCompilerOutputsExtension because classpath is now obtained post compilation instead of at import stage

wisechengyi:
1. Remove PantsCompilerOutputsExtension because classpath is now obtained via target id after export-classpath instead of import stage. 
2. Remove the need to call 'pants compile help-advanced'
3. Other minor refactoring

This RB will slightly improve memory usage and performance.


wisechengyi:
a10543e617c8d0d3c895cd1c9d23a536a5df248c

PR #107, opened 2016-02-04
Remove IntelliJ Compiler option

wisechengyi:
1. There is no reason to use IntelliJ compiler for pants project any more.
2. Having an different compiler for pants project is confusing
3. IntelliJ compiler does not work most of the time for complicated pants projects.


wisechengyi:
9a9cce53923c65af3314c74572ec048377099067

PR #119, opened 2016-04-12
Improve PantsOptions to be more reusable

peiyuwang:
This is a minor clean up, try to reduce the methods added to PantsOptions
by using some more generic functions, has and get.

For example for the new option that I plan to check `test.junit.strict_jvm_version`
Now I can just call `options.get("test.junit.strict_jvm_version")`.

### DIFF
--- a/.idea/misc.xml
+++ b/.idea/misc.xml
@@ -3,7 +3,7 @@
   <component name="EntryPointsManager">
     <entry_points version="2.0" />
   </component>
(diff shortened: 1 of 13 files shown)


peiyuwang:
Reviewed at https://rbcommons.com/s/twitter/r/3685/

Submitted as 7583350692114fe9d4ee921592a3644ba2eecce8

PR #121, opened 2016-04-13
Auto discover and set project SDK for pants plugin

peiyuwang:
This eliminates the manual step to select project SDK, which is

* extra work.
* error prone, given user might select a different sdk from what pants uses.

This is based on https://rbcommons.com/s/twitter/r/3680/ which exports pants
distributions by platforms used. Therefore it will decide based on whether
the new information is available (since export version `1.0.7`) whether to
skip the manual import step.

Other misc changes include:

* Make `bytecodeTargetLevel` 1.7 so IJ is happy with "multi exception catch".
* Fix some old pants argument in the README.md.
* Extract `SimpleExportResult` as a separate class and added tests.


peiyuwang:
Reviewed at https://rbcommons.com/s/twitter/r/3690/

Merged as c85b9593390db042c17bbb5c23a928ea80129df8

PR #123, opened 2016-04-13
Fix view switch thread handle loss

wisechengyi:
The thread handles for view switching are overwritten because multiple projects try to refresh at the same time, so some cannot end because they self loop unless explicitly cancelled. This change blocks any view switch request if there is one ongoing via semaphores.


wisechengyi:
9c8e5820fdd4e1dba727fc953f0ec09c2320f161

PR #124, opened 2016-04-14
plugin work to relay project set up by pants `idea-plugin` goal

wisechengyi:
This is intellij plugin side of work to handle pants `idea-plugin` goal https://rbcommons.com/s/twitter/r/3664/.


wisechengyi:
f3566aba3022844632aa0f0a7ca5ed9cfb8aac4e

PR #125, opened 2016-04-17
Invoke pants regardless sources have changed because resources may have changed.

wisechengyi:
Currently the changes in resources will not be detected by `hasDirtyTargets(holder)` in `PantsTargetBuilder` because it can only detect changes in `PantsBuildTargetType` whereas resources are handled by `ResourcesTargetType`. Therefore this RB disgard any check on `DirtyFilesHolder<PantsSourceRootDescriptor, PantsBuildTarget> holder`, so the behavior will be idential with `PantsCompile` in https://rbcommons.com/s/twitter/r/3607/

It is possible to design another Builder to handle the changes `ResourcesTargetType`, but it will also introduce the issue where both builders will be called without knowing whether the other has performed the job, so the work may be duplicated.

It is also possible to mark resources as sources, but UX will be bad because it will unnecessarily increase the load on already-slow source indexing, and many "syntax errors" will be found.


wisechengyi:
38d0a01d20c8cd86687ed7bfd5713206c8ae1572

PR #132, opened 2016-05-04
Make PantsMakeBeforeRun stateless

wisechengyi:
This RB separates out the following changes from the general cleanup https://rbcommons.com/s/twitter/r/3812/ (comments are also addressed in the RB) 

* Make PantsMakeBeforeRun stateless as suggested by the SDK doc that storing Project may lead to memory leak. https://github.com/JetBrains/intellij-community/blob/10a4c91dbfca935b4b5531003fd8c13b56e66202/plugins/devkit/testData/inspections/statefulEp/Ext.java#L4
* privatize some PantsUtil.findPantsExecutable interfaces.
* Minor `PantsSystemProjectResolver.java` refactoring.


wisechengyi:
bff150a640e6f038295ef2dbbce4566b426670e8

Issue #133, opened 2016-05-05
Compilation failed with error: Abnormal build process termination:


peiyuwang:
In working on #130 , after https://rbcommons.com/s/twitter/r/3819/ ignoring 'leaking' thread in tests, 

Last remaining failure is
```
There was 1 failure:
                     1) testHello(com.twitter.intellij.pants.integration.OSSPantsExamplesMultiTargetsIntegrationTest)
                     junit.framework.AssertionFailedError: Compilation failed with error: Abnormal build process termination: 
[...]

Issue #130, opened 2016-05-03
thread leak error in CI


wisechengyi:
since 2016 change, CI is broken due to thread leak. ticket filed at https://youtrack.jetbrains.com/issue/IDEA-155496

wisechengyi:
ThreadTracker.java has many changes lately. 

Also the check was disabled essentially disabled in IDEA 15.

```
[tw-mbp-yic intellij-community ((pycharm/145.607))]$ git diff master platform/testFramework/src/com/intellij/testFramework/ThreadTracker.java
diff --git a/platform/testFramework/src/com/intellij/testFramework/ThreadTracker.java b/platform/testFramework/src/com/intellij/testFramework/ThreadTracker.java
index 5ed6d1a..b2844e6 100644
--- a/platform/testFramework/src/com/intellij/testFramework/ThreadTracker.java
+++ b/platform/testFramework/src/com/intellij/testFramework/ThreadTracker.java
@@ -15,13 +15,12 @@
  */
 package com.intellij.testFramework;
 
-import com.intellij.execution.process.ProcessIOExecutorService;
 import com.intellij.openapi.Disposable;
 import com.intellij.openapi.application.ApplicationManager;
 import com.intellij.openapi.project.ProjectManager;
 import com.intellij.openapi.project.impl.ProjectManagerImpl;
+import com.intellij.openapi.util.Condition;
 import com.intellij.openapi.util.Disposer;
-import com.intellij.openapi.util.ShutDownTracker;
 import com.intellij.util.ReflectionUtil;
 import com.intellij.util.WaitFor;
 import com.intellij.util.containers.ContainerUtil;
@@ -54,7 +53,7 @@ public class ThreadTracker {
   private static final Method getThreads = ReflectionUtil.getDeclaredMethod(Thread.class, "getThreads");
 
   @NotNull
-  public static Collection<Thread> getThreads() {
+  private static Collection<Thread> getThreads() {
     Thread[] threads;
     try {
       // faster than Thread.getAllStackTraces().keySet()
@@ -66,11 +65,15 @@ public class ThreadTracker {
     return ContainerUtilRt.newArrayList(threads);
   }
 
-  private static final Set<String> wellKnownOffenders = new THashSet<>();
+  private static final Set<String> wellKnownOffenders = new THashSet<String>();
   static {
+        wellKnownOffenders.add("Action Updater"); // todo remove
+        wellKnownOffenders.add("Alarm pool(own)");
[...]

PR #134, opened 2016-05-05
Fix plugin ci: temporarily ignore 'leak' thread and failing test in tests

peiyuwang:
Because of [1] [2], pants plugin CI is red after https://rbcommons.com/s/twitter/r/3780/,
This review fixes broken CI.

* 'leak' in quotes is because it's not clear whether they are real leaks or
  just legit threads forgotten being whitelisted.
* [3] is another failing test that have be disabled, hard to tell if they
  are related to [1] [2], more investgation later.

Detailed changes:

* Assign our own threads a prefix for debugging purpose, so far haven't seen any
  indication the 'leak' is from us.
* Use `ThreadTracker.longRunningThreadCreated` to ignore a few threads we have seen
  in CI, see [4].
* Change `scheduleAtFixedRate` to `scheduleWithFixedDelay`, this does not make a
  difference, but only to be cautious, because of [5]. 
* Ignore failing test [3].
* Added back a few tests not globed.

Next steps:

* For [1] [2], determine the impact, could be a bug we can safely ignore while
  it's being fixed upstream or it can be a 16 launch blocker.
* [3] needs to be addressed asap.

[1] upstream bug: https://youtrack.jetbrains.com/issue/IDEA-155496
[2] plugin bug: https://github.com/pantsbuild/intellij-pants-plugin/issues/130
[3] plugin bug: https://github.com/pantsbuild/intellij-pants-plugin/issues/133
[4] `ThreadTracker` where 'leaking' threads are detected: https://github.com/JetBrains/intellij-community/blob/3abf4b4515f4a91ba3a96326fc1fcc348763e866/platform/testFramework/src/com/intellij/testFramework/ThreadTracker.java#L103
[5] By documentation in IJ code, they try to avoid `scheduleAtFixedRate`: https://github.com/JetBrains/intellij-community/blob/7e16c042a19767d5f548c84f88cc5edd5f9d1721/platform/util/src/com/intellij/util/concurrency/AppExecutorUtil.java


peiyuwang:
Noticed in travis, tests are now quite flaky, saw ivy errors all the time that I have to restart.

peiyuwang:
Reviewed at https://rbcommons.com/s/twitter/r/3819/
Submitted as be0cce2b274743b8bb1e879a90e532494c327d82

Issue #135, opened 2016-05-05
Ivy error when upgrading pants intellij-pants-plugin to 1.0.0


peiyuwang:
Consistently happening, it's not the usual flaky ivy error, the stacktrace isn't super useful to indicate which artifact it has trouble with, it just says _pants.backend.jvm.ivy_utils.IvyResolveResult object at 0x413da10_

Link to [travis](https://travis-ci.org/peiyuwang/intellij-pants-plugin/jobs/128), log is also attached
[ivy-error-log.txt](https://github.com/pantsbuild/intellij-pants-plugin/files/250890/ivy-error-log.txt)

```
17:06:07 00:04       [ivy-resolve]
17:06:13 00:10   [complete]
               FAILURE
Exception caught: (<class 'pants.backend.jvm.ivy_utils.IvyResolveMappingError'>)
  File "/home/travis/.cache/pants/setup/bootstrap/1.0.0/bin/pants", line 11, in <module>
    sys.exit(main())
  File "/home/travis/.cache/pants/setup/bootstrap/pants.Zmuelz/install/local/lib/python2.7/site-packages/pants/bin/pants_exe.py", line 44, in main
    PantsRunner(exiter).run()
  File "/home/travis/.cache/pants/setup/bootstrap/pants.Zmuelz/install/local/lib/python2.7/site-packages/pants/bin/pants_runner.py", line 53, in run
    options_bootstrapper=options_bootstrapper)
  File "/home/travis/.cache/pants/setup/bootstrap/pants.Zmuelz/install/local/lib/python2.7/site-packages/pants/bin/pants_runner.py", line 43, in _run
    return LocalPantsRunner(exiter, args, env, options_bootstrapper=options_bootstrapper).run()
  File "/home/travis/.cache/pants/setup/bootstrap/pants.Zmuelz/install/local/lib/python2.7/site-packages/pants/bin/local_pants_runner.py", line 49, in run
    self._maybe_profiled(self._run)
  File "/home/travis/.cache/pants/setup/bootstrap/pants.Zmuelz/install/local/lib/python2.7/site-packages/pants/bin/local_pants_runner.py", line 46, in _maybe_profiled
    runner()
  File "/home/travis/.cache/pants/setup/bootstrap/pants.Zmuelz/install/local/lib/python2.7/site-packages/pants/bin/local_pants_runner.py", line 89, in _run
    result = goal_runner.run()
[...]

PR #140, opened 2016-05-09
Upgrade pants in intellij-pants-plugin repo to 1.0.0

peiyuwang:
Only issue was https://github.com/pantsbuild/intellij-pants-plugin/issues/135, which
was due to a improper `jar_library.url` syntax that results filename `gson-2.3.1.jar`
being created as a directory.

Verify this has always being the case, but there wasn't any sanity check until
https://rbcommons.com/s/twitter/r/3555/ from release 0.0.77.

This suggests we can also remove `gson-2.3.1` from build file. Other references are
still valid for building artifact purpose.


peiyuwang:
Reviewed at https://rbcommons.com/s/twitter/r/3822/

Submitted as 4660fca0e5bdd0f588e0eb42056b7cf8232aeb6c

PR #146, opened 2016-05-13
Fix GUI stuck focusing on directory/file

wisechengyi:
* replace `queueSwitchToProjectFilesTreeView` with callback API.
* make sure to focus on directory only, rather than a mixture of directories and files.


wisechengyi:
9c838ed8c4a9885ae6ebbd26650c71b9c3a98f9b

PR #156, opened 2016-05-20
Conditionally auto import

peiyuwang:
https://rbcommons.com/s/twitter/r/3690/ doesn't consider python use case. Python projects
will also skip manual selection and get JDK auto configured.

This review attempts to only auto configure project JDK when it's confident enough this is
a java/scala project. It includes

* Revert https://rbcommons.com/s/twitter/r/3816/
* A heuristic based project type detection to determine whether a project is java/scala,
  and only set jdk when that's true.

Follow ups:

* Set project level interpreter for python projects
* Overwrite at module level, jdk for java/scala projects and interpreter for python.


peiyuwang:
Merged as b71c67fe7f2304745ce45d0621dbe60bcccd118c
Reviewed at https://rbcommons.com/s/twitter/r/3856/

PR #157, opened 2016-05-20
cleanup

wisechengyi:
* --no-export-classpath-use-old-naming-style and whether target id is in pants export no longer need to be checked because pants upgrades.
* general cleanup


wisechengyi:
b71c67fe7f2304745ce45d0621dbe60bcccd118c

PR #171, opened 2016-06-13
 Cache SimpleExportResult and PantsOptions based on PantsExecutable; Offer invalidation via GUI; Add Project refresh via GUI

wisechengyi:


wisechengyi:
b8c5939e68b63095293cc9d4603ce206b13eb333

PR #184, opened 2016-07-13
Various fixes for Pants Test configuration

wisechengyi:
Earlier Pants test determines the target after user click `Run` or `Debug` by going up the directories until it finds a BUILD file, then append `::`, so
1. User has no flexibilty selecting targets in the configuration.
2. `::` is not always correct because e.g. `run` goal should only take one target. and Pants test configuration should be designed to handle generic pants commands.

This RB fix them by moving target info explicitly into configuration's script parameters.

Additional fixes:
1. Remove `PantsCompile` from Pants Test configuration, as `PantsCompile` is intended for configuration that needs Pants' product but does not associate with Pants by default.
2. Add `!isGenTarget(targetAdress)` filter on address to compile.


wisechengyi:
bef84b8f106553696c0a6df7adc2605bb6702dfd

Issue #187, opened 2016-07-22
Refactor Pants actions


ebubae:
Current actions (especially compilation actions) reuse a lot of code and are begging for inheritance.

PR #190, opened 2016-07-28
Refactor Pants actions

ebubae:
Address https://github.com/pantsbuild/intellij-pants-plugin/issues/187 and add polymorphism to Pants compile actions.


ebubae:
#190

PR #224, opened 2016-11-23
IntelliJ 2016.3 Compatibility

wisechengyi:
* Set project scala library version explicitly to 2.11.6 in intellij and pants settings because scala plugin is built with 2.11.6
* Add `scalameta` dependency to scala plugin as it now leverages scala meta. 
* Update external-system-test-api.jar from `idea/163.6957.18`
* Remove stale code and run configuration settings


wisechengyi:
Thanks @mateor . Any effort (including dropping by) is appreciated!

PR #225, opened 2016-11-29
Release 1.4.10

wisechengyi:
* IntelliJ 2016.3 Compatibility
* Incompatible with 2016.2 and earlier versions


wisechengyi:
Going to TBR this since there's nothing too exciting about this change except a version bump.

PR #226, opened 2016-11-29
Revert "Release 1.4.10"

wisechengyi:
Reverts pantsbuild/intellij-pants-plugin#225 because it is supposed to be a major version bump for IntelliJ 2016.3


wisechengyi:
Going to TBR this since it is a revert, and nothing else interesting.

PR #228, opened 2016-12-01
Update scala plugin to 2016.3.4

wisechengyi:
Update scala plugin to 2016.3.4


wisechengyi:
Went in with e80cf9415ec5fa7841123b206cb94fa83a40c9f8

PR #229, opened 2016-12-01
Auto deploy master to BleedingEdge channel of the plugin repo

wisechengyi:
Add script to run after success to deploy the plugin to the BleedingEdge channel of https://plugins.jetbrains.com/plugin/7412. Configuring `USERNAME` and `PASSWORD`env is required in Travis CI.

Additional testing: I tried it on my own travis ci and deploy succeeded. https://api.travis-ci.org/jobs/180317145/log.txt?deansi=true


scripts/deploy/deploy.py
```diff
@@ -0,0 +1,85 @@
...
+  sha = get_head_sha()
+  logger.info('Append git sha {} to plugin version'.format(sha))
+
+  tree = ET.parse(PLUGIN_XML)
+  root = tree.getroot()
+
+  # Find the `version` tag then append the head sha to it.
+  version = root.find('version')
+  if version is None:
+    logger.error("version tag not found in %s".format(PLUGIN_XML))
```
baroquebobcat: I think you meant to use `not found in {}`. `%s` won't be substituted by format.

You could skip the format, and use `logger.error('.... %s', ...)` instead.

scripts/deploy/deploy.py
```diff
@@ -0,0 +1,85 @@
...
+    .format(channel=CHANNEL,
+            username=os.environ['USERNAME'],
+            password=os.environ['PASSWORD'],
+            plugin_id=PLUGIN_ID,
+            plugin_jar=PLUGIN_JAR)
+
+  logger.info('Uploading...')
+  try:
+    with open(os.devnull, 'w') as devnull:
+      subprocess.check_output(upload_cmd, shell=True, stderr=devnull)
```
baroquebobcat: If you don't care about the output, it'd make more sense to use `check_call`, or `call`. 

Since this will always return an error according to the comment below, I think it'd be better to just use `call`.

scripts/deploy/deploy.py
```diff
@@ -0,0 +1,85 @@
...
+  except subprocess.CalledProcessError as e:
+    # Plugin upload will return error even if it succeeds,
+    # so the error is meaningless. Need to manually check on the
+    # plugin repo.
+    try:
+      with open(os.devnull, 'w') as devnull:
+        subprocess.check_output('curl {} | grep {}'.format(REPO, sha), shell=True, stderr=devnull)
```
baroquebobcat: You could just use `call`, then you wouldn't need the `try`. Instead, you could use a conditional on the result.

scripts/get_3rdparty_jars.py
```diff
@@ -0,0 +1,8 @@
+import subprocess
+
+output = subprocess.check_output("./pants dependencies : 3rdparty/intellij:: | sort | grep 'com\|org\|jdk'", shell=True)
```
baroquebobcat: Interesting. I wouldn't have expected `:` by itself to be a valid commandline address.

scripts/deploy/deploy.py
wisechengyi: corrected to `{}` for consistency

scripts/deploy/deploy.py
wisechengyi: used `call` instead.
thank you for the reasoning!

scripts/get_3rdparty_jars.py
wisechengyi: yeah it is equivalent to `//:` since the script is invoked from buildroot

PR #230, opened 2016-12-07
Added support for loading minimized roots 

dbrewster:
using the technique of:
  for target in $(./pants minimize ::)
    m = $(./pants export target)
    merge M into ProjectInfo

### DIFF
--- a/src/com/twitter/intellij/pants/service/PantsCompileOptionsExecutor.java
+++ b/src/com/twitter/intellij/pants/service/PantsCompileOptionsExecutor.java
@@ -28,9 +28,12 @@ import org.jetbrains.annotations.TestOnly;
 
 import java.io.File;
 import java.io.IOException;
+import java.util.Arrays;
+import java.util.Collections;
 import java.util.List;
 import java.util.Optional;
 import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 public class PantsCompileOptionsExecutor {
   protected static final Logger LOG = Logger.getInstance(PantsCompileOptionsExecutor.class);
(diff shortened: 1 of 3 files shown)


src/com/twitter/intellij/pants/service/PantsCompileOptionsExecutor.java
```diff
@@ -131,15 +134,51 @@ public PantsCompileOptions getOptions() {
...
+  private String[] loadMinimizedTargets(
+    @NotNull Consumer<String> statusConsumer,
+    @Nullable ProcessAdapter processAdapter
+  ) throws IOException, ExecutionException {
+    statusConsumer.consume("Determining optimal import set...");
+    final File outputFile = FileUtil.createTempFile("pants_minimzie_run", ".out");
+    final GeneralCommandLine commandLine = PantsUtil.defaultCommandLine(getProjectPath());
+    commandLine.addParameter("minimize");
+    commandLine.addParameter("--minimize-output-file=" + outputFile.getPath());
+    commandLine.addParameter("::");
```
wisechengyi: still need to use `getAllTargetAddresses` instead of hardcoding `::`, otherwise when user selects `buildroot/a/b/c::` it will still do `buildroot::`

P.S. `getAllTargetAddresses` really means getting the target specs, one of the legacies I haven't got to yet.

src/com/twitter/intellij/pants/service/PantsCompileOptionsExecutor.java
```diff
@@ -195,6 +234,41 @@ private ProcessOutput getProcessOutput(
   }
 
   @NotNull
+  private String getExport(
+    @NotNull String target,
+    @Nullable ProcessAdapter processAdapter
+  ) {
+    try {
+      final File outputFile = FileUtil.createTempFile("pants_export_" + target.replace('/', '_') + "_run", ".out");
```
wisechengyi: It's okay to have a static output file name, since it will be loaded before the next pants run and the next pants run will overwrite it.

If you really want to have a different name for each run, please see use `File.separator` instead of `/`. See `com.twitter.intellij.pants.service.PantsCompileOptionsExecutor#getProjectName` for example.

src/com/twitter/intellij/pants/service/project/PantsResolver.java
```diff
@@ -80,7 +89,7 @@ public void resolve(
     @Nullable ProcessAdapter processAdapter
   ) {
     try {
-      String pantsExportResult = myExecutor.loadProjectStructure(statusConsumer, processAdapter);
+      List<String> pantsExportResult = myExecutor.loadProjectStructure(statusConsumer, processAdapter);
```
wisechengyi: `pantsExportResults`

src/com/twitter/intellij/pants/service/project/model/ProjectInfo.java
```diff
@@ -180,4 +180,12 @@ public String toString() {
            ", targets=" + targets +
            '}';
   }
+
+  public void merge(ProjectInfo projectInfo) {
+    getLibraries().putAll(projectInfo.getLibraries());
+    getTargets().putAll(projectInfo.getTargets());
```
wisechengyi: I'd prefer to use the instance variables directly get as an indicator it is modifying something within the class.

`getLibraries()` and `getTargets()` really mean to return a frozen collection. Sorry legacy issue again.

src/com/twitter/intellij/pants/service/project/model/ProjectInfo.java
```diff
@@ -182,10 +182,10 @@ public String toString() {
   }
 
   public void merge(ProjectInfo projectInfo) {
-    getLibraries().putAll(projectInfo.getLibraries());
-    getTargets().putAll(projectInfo.getTargets());
-    if (getPythonSetup() == null && projectInfo.getPythonSetup() != null) {
-      python_setup = projectInfo.getPythonSetup();
+    libraries.putAll(projectInfo.libraries);
+    targets.putAll(projectInfo.targets);
```
wisechengyi: This should be something like 
`libraries.putAll(projectInfo.getLibraries())` as `libraries` is an internal variable to we want to modify, `projectInfo` is from "outside" which we just want to read only from.

wisechengyi:
For test data, you can follow the example in testData/dummy_repo

The test framework copies from dummy_repo to the test root, then try to import on it, like the following:
https://github.com/pantsbuild/intellij-pants-plugin/search?utf8=%E2%9C%93&q=dummy_repo

Alternatively you can add examples directly under dummy_repo to show the difference of 3rdparty resolves. For example:
With the existing import scheme
1. import some project
2. assert some module
3. then assert certain libraries under that module

With the new scheme
1. same
2. same or assert a different module
3. assert different sets of libraries in that module

wisechengyi:
Closing this to due inactivity.

PR #232, opened 2016-12-13
Update scala plugin

wisechengyi:
Update scala plugin


wisechengyi:
TBR since there's nothing interesting.

PR #237, opened 2016-12-27
IntelliJ 2017 Support

wisechengyi:
## Changes

* IJ 2017 compatibility and backward **incompatibility** with 2016.3 due to python plugin api breakages
* Adjusting the mechanism to collect python plugin jars accordingly because community version's name has been changed to `python-ce` and ultimate version remains as `python`.
* Explicitly add gradle plugin as a dependency because pants runner uses it as a library
* Regenerated the list to exclude when building the final distributable pants support zip.
* Allow 3rdparty jar collection to specify their own namespace `com.intellij.<plugin_name>` in order to avoid conflicting cases.


wisechengyi:
currently blocked by https://youtrack.jetbrains.com/issue/IDEA-169233

PR #242, opened 2017-01-10
Move Pants output to console view instead of notification channel

wisechengyi:
## Problem
Pants output used to go out via the IntelliJ notification channel with a few down sides:
1. No auto scrolling supported. If the message gets long, user has to manually scroll the message window.
2. Output does not show up exactly like the command line. E.g. compile error indicator `^` will show up off.
3. Cannot take advantage of IntelliJ hyperlink for user to click on the error to directly dump to what the error stems.

screen shot 2017-01-11 at 11 13 21 am

## Solution
Use a text console view, which has
1. Auto scroll by default
2. Correct output format
3. Capability to inject hyperlinks

screen shot 2017-01-11 at 11 18 58 am


tests/com/twitter/intellij/pants/execution/PantsMakeMessageTest.java
```diff
@@ -0,0 +1,53 @@
...
+    assertEquals(
+      "/Users/dummy/workspace/pants_ij/examples/tests/java/org/pantsbuild/example/hello/greet tv/GreetingTest.java",
+      result.get().getFilePath()
+    );
+    assertEquals(23, result.get().getLineNumber());
+    assertEquals(1, result.get().getColumnNumber());
+  }
+
+
+  public void testErrorMessageWithLocation2() {
```
peiyuwang: what's the difference from the java test case? seems redundant

src/com/twitter/intellij/pants/execution/PantsMakeBeforeRun.java
```diff
@@ -348,9 +329,124 @@ else if (output.contains("[error]")) {
...
+    }
+
+    String[] splitByColon = line.split(":");
+    if (splitByColon.length < 3) {
+      return Optional.empty();
+    }
+
+    try {
+      // filePath path is between tag and first colon
+      String filePath = splitByColon[0].substring(splitByColon[0].indexOf(tag) + tag.length()).trim();
```
peiyuwang: some negative examples from old log
```
[error] Compile failed at Dec 9, 2016 6:33:30 AM [0.080s]
[error]         Assert.assertEquals("0:00:00.000", ManagementUtils.getAverageAge(d0));
```

wisechengyi: It's okay for `String filePath` to be `Assert.assertEquals("0` for example, because [the hyperlink construction will check if it is an actual file](https://github.com/pantsbuild/intellij-pants-plugin/pull/242/files/c3d1080f4acc15027ff7b6d3b505d8ee7c693dc5#diff-77d87e704dfb9b6b30ffaa3a11f7d6cdR343). If not, it will there will be no hyperlinks. 

I will add more doc to this to be more clear.

tests/com/twitter/intellij/pants/execution/PantsMakeMessageTest.java
wisechengyi: Thanks! I will remove having certain file extensions as a requirement, because eventually we will check if it is a valid file path when constructing the hyperlink.

src/com/twitter/intellij/pants/execution/PantsMakeBeforeRun.java
wisechengyi: More doc added, as well as test case to cover this.

src/com/twitter/intellij/pants/execution/PantsMakeBeforeRun.java
```diff
@@ -348,9 +327,121 @@ else if (output.contains("[error]")) {
...
+      @Nullable
+      @Override
+      public Result applyFilter(String line, int entireLength) {
+        Optional<ParseResult> result = ParseResult.parseErrorLocation(line, ERROR_TAG);
+        if (result.isPresent()) {
+
+          OpenFileHyperlinkInfo linkInfo = new OpenFileHyperlinkInfo(
+            project,
+            result.get().getFile(),
+            result.get().lineNumber - 1, // line number needs to be 0 indexed
```
peiyuwang: `getLineNumber` `getColumnNumber`

peiyuwang: q: in case we have a invalid line number either `<0` or `> max actual line#` what will happen? as long as it doesn't crash.

wisechengyi: 1. corrected.
2. Manually tested by intentionally feeding it with line number -1, there is no crash, just the cursor is somewhere random.

wisechengyi: 3. For non digits will be caught by `NumberFormatException`, and return no link.

peiyuwang: 👍

wisechengyi:
Thanks @peiyuwang !

PR #249, opened 2017-01-20
Langugage level fix

wisechengyi:
## Problem
For a fresh IJ environment, the JDK pants plugin selected cannot be automatically recognized in terms of language level.

## Solution
* Adjust the jdkNames to be `1.8_xxxx` so intellij can automatically pick up the language level
* Remove hardcoded project sdk in test framework and use the jdk created via pants instead, in order to verify the fix works.

## Other change
* Remove pants 1.1.0 from `.travis.yml` since it was old and still uses java6 as default.
* Add 1.2.1

### DIFF
--- a/common/com/twitter/intellij/pants/util/PantsUtil.java
+++ b/common/com/twitter/intellij/pants/util/PantsUtil.java
@@ -780,6 +780,7 @@ public class PantsUtil {
     if (!jdkHome.isPresent()) {
       return Optional.empty();
     }
+
     Sdk jdk = JavaSdk.getInstance().createJdk(jdkName, jdkHome.get());
     return Optional.of(jdk);
   }
--- a/common/com/twitter/intellij/pants/util/PantsUtil.java
+++ b/common/com/twitter/intellij/pants/util/PantsUtil.java
@@ -768,7 +768,7 @@ public class PantsUtil {
     String jdkName = "1.x_from_pants";
     for (String version : ContainerUtil.newArrayList("6", "7", "8", "9")) {
       if (defaultPlatform.contains(version)) {
-        jdkName = String.format("1.%s_from_pants", version);
+        jdkName = String.format("1.%s_from_%s", version, pantsExecutable);
         break;
       }
     }
--- a/common/com/twitter/intellij/pants/util/PantsUtil.java
+++ b/common/com/twitter/intellij/pants/util/PantsUtil.java
@@ -763,7 +763,15 @@ public class PantsUtil {
     }
 
     final String defaultPlatform = exportResult.getJvmPlatforms().getDefaultPlatform();
-    final String jdkName = String.format("JDK from pants %s", defaultPlatform);
+
+    // if defaultPlatform is java8, then jdkName should be 1.8_from_pants
+    String jdkName = "1.x_from_pants";
+    for (String version : ContainerUtil.newArrayList("6", "7", "8", "9")) {
+      if (defaultPlatform.contains(version)) {
+        jdkName = String.format("1.%s_from_pants", version);
+        break;
+      }
+    }
 
     boolean strict = PantsOptions.getPantsOptions(pantsExecutable)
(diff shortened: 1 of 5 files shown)


common/com/twitter/intellij/pants/util/PantsUtil.java
```diff
@@ -763,7 +763,15 @@ public static boolean supportExportDefaultJavaSdk(@NotNull final String pantsExe
     }
 
     final String defaultPlatform = exportResult.getJvmPlatforms().getDefaultPlatform();
-    final String jdkName = String.format("JDK from pants %s", defaultPlatform);
+
+    // if defaultPlatform is java8, then jdkName should be 1.8_from_pants
+    String jdkName = "1.x_from_pants";
```
baroquebobcat: It would be useful to still include the jdkName in this default name. That way if the loop below didn't match, the name would have the platform info in it.

common/com/twitter/intellij/pants/util/PantsUtil.java
```diff
@@ -763,7 +763,15 @@ public static boolean supportExportDefaultJavaSdk(@NotNull final String pantsExe
...
+    for (String version : ContainerUtil.newArrayList("6", "7", "8", "9")) {
+      if (defaultPlatform.contains(version)) {
```
baroquebobcat: I could imagine circumstances where a platform name is something like jdk_1.9.0_u66. For example, if a repo is testing with different jdk patch releases. Admittedly, a name like that goes against the convention in the pants repo, but it isn't disallowed.
If this checked against the source level rather than the name, it would not be vulnerable to this. And, it's behavior would align better with the meaning of the platform config.

testFramework/com/twitter/intellij/pants/testFramework/OSSPantsIntegrationTest.java
```diff
@@ -80,4 +84,21 @@ protected void assertNotContainsSubstring(String s, String unexpected) {
...
+   * Assert Project has the right JDK and language level (JVM project only).
+   */
+  protected void assertProjectJdkAndLanguageLevel() {
+    assertEquals(
+      ProjectRootManager.getInstance(myProject).getProjectSdk().getHomePath(),
+      PantsUtil.getDefaultJavaSdk(getParentPath()).get().getHomePath()
+    );
+
+    LanguageLevel projectLanguageLevel = LanguageLevelProjectExtension.getInstance(myProject).getLanguageLevel();
+    LanguageLevel expectedLanguageLevel = LanguageLevel.JDK_1_8;
```
baroquebobcat: I'd like to see a test where the language level default is not 1.8.

common/com/twitter/intellij/pants/util/PantsUtil.java
wisechengyi: do you mean to have the jdkName default to 
`jdkName = String.format("1.x_from_%s", pantsExecutable)`?

common/com/twitter/intellij/pants/util/PantsUtil.java
wisechengyi: Switched to using IJ's framework to detect jdk version. so jdkInfo.getVersion() returns `java version "1.8.0_121"` then we can check for `1.7` `1.8` `1.9`

wisechengyi:
Please see #251

testFramework/com/twitter/intellij/pants/testFramework/OSSPantsIntegrationTest.java
wisechengyi: This requires my laptop and travis to have different java version installed, which is a bit hard to accommodate. Since I did not explicitly set language level anywhere and the jdk level was determined via IntelliJ framework's library, I'd imagine the existing test should suffice.

baroquebobcat: I figured it was something like that. Thanks for the follow up!

PR #251, opened 2017-01-26
Language level fix - follow up

wisechengyi:
Follow up for #249


wisechengyi:
Thanks for the follow up as well!

PR #261, opened 2017-02-07
Skip looking for JDK when there is one set already

wisechengyi:
Other change:
Remove Pants export version >= 1.0.7 because it's always true.


wisechengyi:
Test added. Regarding version check, 1.0.7 was added Apr 2016, so it's safe to assume everything in on or above it.

wisechengyi:
Hi @baroquebobcat, would you like to make additional comment? I plan to merge this by EOD.

PR #298, opened 2017-05-22
Add compile current target action

wisechengyi:
This change adds a quick way to trigger Pants compile for the target under edit. User can choose to tie the action in their keymap.

### DIFF
--- a/src/com/twitter/intellij/pants/compiler/actions/PantsCompileCurrentTargetAction.java
+++ b/src/com/twitter/intellij/pants/compiler/actions/PantsCompileCurrentTargetAction.java
@@ -4,6 +4,8 @@
 package com.twitter.intellij.pants.compiler.actions;
 
 import com.intellij.openapi.actionSystem.AnActionEvent;
+import com.intellij.openapi.editor.Editor;
+import com.intellij.openapi.editor.impl.EditorImpl;
 import com.intellij.openapi.fileEditor.FileEditorManager;
 import com.intellij.openapi.module.Module;
 import com.intellij.openapi.project.Project;
@@ -12,10 +14,6 @@ import com.intellij.openapi.vfs.VirtualFile;
 import com.twitter.intellij.pants.util.PantsUtil;
 import org.jetbrains.annotations.NotNull;
 
-import java.util.Arrays;
-import java.util.Collection;
-import java.util.Set;
-import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
 /**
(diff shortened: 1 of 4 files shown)


tests/com/twitter/intellij/pants/integration/OSSPantsCompileActionsTest.java
```diff
@@ -77,6 +86,36 @@ public void testCompileAllTargetsInModuleAction() throws Throwable {
...
+      String target = classAndTarget.getSecond();
+
+      PsiClass clazz = JavaPsiFacade.getInstance(myProject)
+        .findClass(clazzName, GlobalSearchScope.projectScope(myProject));
+
+      assertNotNull(String.format("%s does not exist, but should.", clazz), clazz);
+
+      FileEditorManager.getInstance(myProject).openFile(clazz.getContainingFile().getVirtualFile(), true);
+      PantsCompileCurrentTargetAction compileCurrentTargetAction = new PantsCompileCurrentTargetAction();
+      Set<String> currentTargets = compileCurrentTargetAction.getTargets(getPantsActionEvent(), myProject).collect(Collectors.toSet());
```
baroquebobcat: Could you adjust the whitespace here a bit to make it clearer that the `getTargets` call is what is exercising the code being tested?

src/com/twitter/intellij/pants/compiler/actions/PantsCompileCurrentTargetAction.java
```diff
@@ -0,0 +1,50 @@
...
+  protected PantsCompileCurrentTargetAction(String name) {
+    super(name);
+  }
+
+  public PantsCompileCurrentTargetAction() {
+    this("Compile target(s) in the selected editor");
+  }
+
+  /**
+   * Find the target(s) that are only associated with the file opened in the selected editor.
```
baroquebobcat: Based on my reading of the javadoc for FileEditorManager, this will currently run compile for all of the files selected in editor panes. Maybe it should only compile the currently focused one?

tests/com/twitter/intellij/pants/integration/OSSPantsCompileActionsTest.java
```diff
@@ -77,6 +86,36 @@ public void testCompileAllTargetsInModuleAction() throws Throwable {
     assertFalse(compileAllTargetsInModuleAction.doCleanAll());
   }
 
+  public void testCompileTargetsInSelectedEditor() throws Throwable {
+    doImport("examples/tests/scala/org/pantsbuild/example");
```
baroquebobcat: Could you add another test that covers what happens if there are multiple editor panes open?

src/com/twitter/intellij/pants/execution/PantsMakeBeforeRun.java
```diff
@@ -199,6 +199,7 @@ public boolean executeTask(
     // If project has not changed since last Compile, return immediately.
     if (!FileChangeTracker.shouldRecompileThenReset(currentProject, targetAddressesToCompile)) {
       PantsExternalMetricsListenerManager.getInstance().logIsPantsNoopCompile(true);
+      showPantsMakeTaskMessage("Already up to date.\n", ConsoleViewContentType.SYSTEM_OUTPUT, currentProject);
```
baroquebobcat: Could you move this to another review? It seems unrelated.

tests/com/twitter/intellij/pants/integration/OSSPantsCompileActionsTest.java
wisechengyi: there are two tests classes and their associated targets, so when the 2nd one is opened, there will be 2 files opened in the editor, but one is selected. I can add a comment to it.

src/com/twitter/intellij/pants/compiler/actions/PantsCompileCurrentTargetAction.java
wisechengyi: Ah I see. I split the editor panes, then it will return targets associated with both files in editor. Let me see if there is a way to find the focused one.

src/com/twitter/intellij/pants/execution/PantsMakeBeforeRun.java
wisechengyi: It is related, because if user triggers the same action without any change, the IDE will noop, which makes a bit confusing.

wisechengyi: will change it to a notification since printing out in console may not be obvious either.

tests/com/twitter/intellij/pants/integration/OSSPantsCompileActionsTest.java
wisechengyi: corrected.

src/com/twitter/intellij/pants/compiler/actions/PantsCompileCurrentTargetAction.java
wisechengyi: corrected to the editor in focus.

PR #303, opened 2017-07-19
1.7.0 Release

dotordogh:
Changes:

- A new feature to compile a target that is being edited. 
- Fix deprecations introduced in IntelliJ 2017.2.
- Refactoring code.
- Updating documentation.

### DIFF
--- a/resources/META-INF/plugin.xml
+++ b/resources/META-INF/plugin.xml
@@ -16,7 +16,7 @@
   <vendor>Twitter, Inc.</vendor>
 
   <!--if you are changing since-build don't forget to change it in .travis.yml file as well-->
-  <idea-version since-build="172.*" until-build="173.0"/>
+  <idea-version since-build="172.0" until-build="172.*"/>
 
   <!--Scala plugin related extensions-->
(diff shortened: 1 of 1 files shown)


resources/META-INF/plugin.xml
```diff
@@ -16,7 +16,7 @@
   <vendor>Twitter, Inc.</vendor>
 
   <!--if you are changing since-build don't forget to change it in .travis.yml file as well-->
-  <idea-version since-build="172.1657" until-build="173.0"/>
+  <idea-version since-build="172.*" until-build="173.0"/>
```
wisechengyi: hmm maybe intellij doesn't take wildcard anymore...try 172.0 instead

wisechengyi: ah looks like `since-build` cannot be wildcard, `until-build` can, so
```
  <idea-version since-build="172.0" until-build="172.*"/>
```
should do.

wisechengyi: example http://www.jetbrains.org/intellij/sdk/docs/basics/getting_started/build_number_ranges.html

PR #304, opened 2017-07-20
Exclude trilead-ssh2 from plugin dist

wisechengyi:
Generated by `scripts/get_3rdparty_jars.py`


wisechengyi:
cc @dotordogh since I cannot add you as a reviewer

wisechengyi:
The purpose of this is exclude the libraries that are already available in the intellij core framework, so we don't need to include it the jar we distribute on the plugin repo.

wisechengyi:
correct solution is at https://github.com/pantsbuild/intellij-pants-plugin/pull/305

PR #305, opened 2017-07-20
Specify scope for 3rdparty java libraries

wisechengyi:
# Problem
Before we have to manually exclude all the 3rdparty jars that will be present at intellij runtime, including the scala runtime which will be supplied by the scala plugin.

# Solution
Mark those 3rdparty deps with `scope='compile test'`, so `./pants binary scripts/sdk:intellij-pants-plugin-publish` will have the thin jar.


wisechengyi:
cc @dotordogh since I cannot add you as reviewer

PR #312, opened 2017-09-07
add lint action

cosmicexplorer:
# Summary
- add item on menu bar and context menu to perform pants lint task on selected file (`PantsLintCurrentTargetAction`)
- restructure menu actions to subclass new file `PantsTaskActionBase`, which uses the appropriate instances of `IPantsGetTargets` and `IPantsExecuteTask` to determine the action's behavior
- restructure `PantsMakeBeforeRun` to support non-compilation tasks as well

# TODO
- local tests are failing and taking forever to run, need to figure out why
- haven't used the action and other actions affected very much yet, need to see if there are any subtle errors
- try to shrink the diff
- the assertions of the result of `.doCleanAll()` have been removed, since that method was removed. do we still want to check somehow whether a `clean-all` was performed to replace that assertion?
- add tests for behavior changed by this diff


wisechengyi:
need to update the scala plugin sha in `./scripts/prepare-ci-environment.sh `, there has been changes upstream.

cosmicexplorer:
Forgot to bring that change up earlier, had kept it unstaged in local worktree. Same test errors when trying to run previously as before, so CI will probably still fail.

src/com/twitter/intellij/pants/compiler/actions/PantsCompileActionGroup.java
```diff
@@ -30,40 +32,43 @@
...
 
-    //  Adds compile all option for modules with multiple targets.
     if (targetAddresses.size() > 1) {
       actions.add(new PantsCompileAllTargetsInModuleAction(module));
     }
-
-    for (String targetAddress : targetAddresses) {
-      actions.add(new PantsCompileTargetAction(targetAddress));
-    }
+    targetAddresses.forEach(target -> actions.add(new PantsCompileTargetAction(target)));
```
dotordogh: Is this change necessary? What benefit does it provide?

src/com/twitter/intellij/pants/execution/PantsMakeBeforeRun.java
```diff
@@ -172,101 +181,127 @@ public boolean executeTask(
...
-    if (!pantsExecutable.isPresent()) {
-      return Pair.create(
-        false,
-        Optional.of(
-          PantsBundle.message("pants.error.no.pants.executable.by.path", currentProject.getProjectFilePath())
-        )
-      );
-    }
-    final GeneralCommandLine commandLine = PantsUtil.defaultCommandLine(pantsExecutable.get().getPath());
+    List<String> pantsCmdLine = new ArrayList<>();
```
wisechengyi: nit: rename var and use simpler api for array, `List<String> pantsCmdArgs = Lists.newArrayList()`

src/com/twitter/intellij/pants/execution/PantsMakeBeforeRun.java
```diff
@@ -163,6 +163,15 @@ public boolean canExecuteTask(
     return true;
   }
 
+  public static class PantsExecuteTaskResult {
```
wisechengyi: How about moving this into a separate file as a standalone class? because it is a bit awkward for a lot of the other class to depend on an inner class e.g.
```
protected PantsMakeBeforeRun.PantsExecuteTaskResult pantsCompileProject() {
...
}
```

src/com/twitter/intellij/pants/execution/PantsMakeBeforeRun.java
```diff
@@ -279,35 +314,21 @@ public void onTextAvailable(ProcessEvent event, Key outputType) {
...
-  /* Show pop up notification about pants compile result. */
+  private static void notify(final String title, final String subtitle, NotificationType type) {
+    /* Show pop up notification about pants compile result. */
     ApplicationManager.getApplication().invokeLater(new Runnable() {
       @Override
       public void run() {
-        String message = success ? "Pants compile succeeded." : "Pants compile failed.";
-        NotificationType type = success ? NotificationType.INFORMATION : NotificationType.ERROR;
-        Notification start = new Notification(PantsConstants.PANTS, "Compile message", message, type);
+        Notification start = new Notification(PantsConstants.PANTS, PantsIcons.Icon, title, subtitle, null, type, null);
```
wisechengyi: any reason for using a different api for `new Notification`?

src/com/twitter/intellij/pants/compiler/actions/PantsGetModuleTargets.java
```diff
@@ -0,0 +1,46 @@
...
+
+public class PantsGetModuleTargets implements IPantsGetTargets {
+
+  private final Optional<Module> module;
+
+  public PantsGetModuleTargets() {
+    this(Optional.empty());
+  }
+
+  public PantsGetModuleTargets(@Nullable Module module) {
```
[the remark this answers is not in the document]
cosmicexplorer: removed

src/com/twitter/intellij/pants/compiler/actions/PantsGetModuleTargets.java
```diff
@@ -0,0 +1,46 @@
...
+
+import java.util.LinkedList;
+import java.util.Optional;
+import java.util.stream.Stream;
+
+public class PantsGetModuleTargets implements IPantsGetTargets {
+
+  private final Optional<Module> module;
+
+  public PantsGetModuleTargets() {
```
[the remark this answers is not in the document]
cosmicexplorer: removed

src/com/twitter/intellij/pants/compiler/actions/IPantsGetTargets.java
```diff
@@ -0,0 +1,34 @@
...
+import java.util.Optional;
+import java.util.stream.Stream;
+
+public interface IPantsGetTargets {
+  static Optional<VirtualFile> getFileForEvent(@Nullable AnActionEvent e) {
+    return Optional.ofNullable(e)
+      .flatMap(ev -> Optional.ofNullable(ev.getData(CommonDataKeys.VIRTUAL_FILE)));
+  }
+
+  static Optional<VirtualFile> getSelectedFile(@Nullable Project project) {
```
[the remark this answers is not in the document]
cosmicexplorer: 1. renamed to `getFileInSelectedEditor`
2. both moved to `PantsUtil`

src/com/twitter/intellij/pants/execution/PantsMakeBeforeRun.java
cosmicexplorer: moved into a standalone class

src/com/twitter/intellij/pants/execution/PantsMakeBeforeRun.java
```diff
@@ -172,101 +181,127 @@ public boolean executeTask(
...
-  /**
-   * @param currentProject:           current project
-   * @param targetAddressesToCompile: set of target addresses to compile
-   * @param useCleanAll:              whether to use clean-all first in Pants
-   * @return whether the execution is successful, additional message along with the execution
-   * in a Pair object.
-   */
-  public Pair<Boolean, Optional<String>> executeTask(Project currentProject, Set<String> targetAddressesToCompile, boolean useCleanAll) {
+  @NotNull
+  public PantsExecuteTaskResult executeCompileTask(@NotNull Project currentProject, @NotNull Set<String> targetAddressesToCompile, @NotNull List<String> tasks) {
```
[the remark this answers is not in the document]
cosmicexplorer: no, that was a mistake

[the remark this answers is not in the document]
cosmicexplorer: No, that was a mistake and I have put it back.

src/com/twitter/intellij/pants/execution/PantsMakeBeforeRun.java
cosmicexplorer: Done.

src/com/twitter/intellij/pants/execution/PantsMakeBeforeRun.java
cosmicexplorer: `PantsMakeBeforeRun` uses `ApplicationManager` here to invoke the notification later (which is what it was doing before, I just put that into this private `.notify()` method). I was 
thinking about making `.notify()` public and call that in `PantsTaskActionBase`, which is the other place a notification is created in this diff, but that might cause weirdness with integration tests since `ApplicationManager` is used here, so I left the notification creation code in two different places.

src/com/twitter/intellij/pants/compiler/actions/PantsLintCurrentTargetAction.java
```diff
@@ -0,0 +1,16 @@
...
+
+/**
+ * PantsLintCurrentTargetAction is a UI action that lints target(s) related to the file under edit.
+ */
+public class PantsLintCurrentTargetAction extends PantsTaskActionBase {
+
+  public PantsLintCurrentTargetAction() {
+    super(new PantsGetJustFileTarget(),
+          new PantsExecuteLint(),
+          "Lint target(s) in the selected editor");
```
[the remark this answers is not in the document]
cosmicexplorer: Fixed, now it invokes the same target list as `PantsCompileAllTargetsInModuleAction`.

cosmicexplorer:
Previous tests all seem to pass here after merging #314 from master. I'm going to add a couple tests for the changes added in this pr -- one to see if the lint command works as expected, another for the changes I made to `PantsContants` and `PantsOptions`.

common/com/twitter/intellij/pants/util/PantsUtil.java
```diff
@@ -222,6 +230,24 @@ public boolean value(VirtualFile virtualFile) {
     ));
   }
 
+  // TODO: when is an Editor not an EditorImpl? why is this functionality not exposed in the intellij openapi Editor interface?
```
wisechengyi: you can browse through the subclasses of `Editor` to get an idea.

common/com/twitter/intellij/pants/util/PantsUtil.java
```diff
@@ -950,4 +976,10 @@ public static void invalidatePluginCaches() {
   public static List<String> parseCmdParameters(Optional<String> cmdArgsLine) {
     return cmdArgsLine.map(ParametersListUtil::parse).orElse(ContainerUtil.newArrayList());
   }
+
+  public static <T, U> Set<U> lookupKeys(Set<T> keys, ImmutableMap<T, ImmutableSet<U>> map) {
```
wisechengyi: Could you comment and test for this? It is unclear to me what it tries to do.

src/com/twitter/intellij/pants/compiler/actions/PantsCompileActionGroup.java
```diff
@@ -30,40 +32,45 @@
...
 
+    List<String> targetAddresses = PantsUtil.getNonGenTargetAddresses(module);
+    // TODO: signal if no addresses found?
     if (targetAddresses.isEmpty()) {
       return emptyAction;
     }
 
-    LinkedList<AnAction> actions = new LinkedList<>();
-
-    //  Adds compile all option for modules with multiple targets.
```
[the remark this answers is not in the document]
cosmicexplorer: I don't think this was intentional -- I'll try to make sure comments and docstrings stay put in the future. I added this back.

common/com/twitter/intellij/pants/util/PantsUtil.java
cosmicexplorer: This was meant to be a note to myself and has been removed. Thanks for the tip.

common/com/twitter/intellij/pants/util/PantsUtil.java
cosmicexplorer: Yes. This bit (along with the changes in `PantsConstants` and `PantsMakeBeforeRun::executeTask`) was written so that the right options get added to the command line for the requested set of tasks. For example, adding `--async` to the command line causes pants to error out unless `clean-all` is one of the tasks. This wasn't an issue before when we were only compiling because we could explicitly check whether we were also running `clean-all`. I'll add comments and tests for this so it's more clear.

.travis.yml
```diff
@@ -42,7 +42,7 @@ env:
     - PANTS_SHA="release_1.2.0" TEST_SET=jvm-integration
 
 script:
-  - ./scripts/run-tests-ci.sh
+  - ./scripts/run-tests-ci.sh --test-junit-output-mode=ALL
```
wisechengyi: revert. you can use PANTS_TEST_JUNIT_OUTPUT_MODE=ALL in travis settings to do the same.

cosmicexplorer: done, left a comment to that effect on the commit

PR #314, opened 2017-09-08
improve exception messages used within import dialog

cosmicexplorer:
- Provide better exception messages when `PantsUtil.listAllTargets()` fails.
  - e.g. whether the process finished with an exit code or was interrupted
- Log the exit code and output file if the process ran to completion.

Currently, the function throws a `PantsException` on failure which describes the command line that failed. This command line already contains the output file, incidentally, but this change makes the output file explicit.

PR #316, opened 2017-10-02
update ij version and checksums

cosmicexplorer:
master currently fails the test suite, because jetbrains has removed the previous version of IDEA for download from their site. This pull request makes the changes listed below so master can run again. This issue was blocking #312 and #314.

- bump ij version to the [most recent available release](https://www.jetbrains.com/idea/download)
  - update list of jars read from the scala plugin directory for the new version of ij
- update checksums for IDEA and plugins
- update scala sdk version to 2.11.11
- explicitly specify python and jdk version in `.travis.yml`
- refactor some calls that were marked deprecated in between previous and current versions of ij


.travis.yml
```diff
@@ -15,6 +14,20 @@ notifications:
   email:
     - yic@twitter.com
```
wisechengyi: non blocking. feel free to add your email if you wish.

cosmicexplorer: Absolutely, I wasn't sure what the protocol was for that.

wisechengyi:
will merge once ci is green

PR #318, opened 2017-10-14
refactor pants invocation to only add cli options when appropriate

cosmicexplorer:
The previous logic adds cli options (e.g. `--async`) to the command line if the version of pants supports e.g. `clean-all.async`, regardless of whether the command line contains the `clean-all` task. This makes it harder to construct different command lines with different tasks such as lint, because if `--async` is added without the `clean-all` task, pants throws an error. Instead, we ensure that cli options such as `--async` are only added if they are relevant to one of the tasks.

An alternative would be to make sure the cli options are less broadly scoped (e.g. `--clean-all-async` instead of `--async`), but adding a bunch of options that either go unused or interfere with the desired result sounds like it could cause problems later.


src/com/twitter/intellij/pants/execution/PantsMakeBeforeRun.java
```diff
@@ -173,46 +175,57 @@ public boolean executeTask(
...
-      Notification start = new Notification(PantsConstants.PANTS, "Pants Compile", "Already up to date.\n", NotificationType.INFORMATION);
-      Notifications.Bus.notify(start);
-      return Pair.create(true, Optional.of(PantsConstants.NOOP_COMPILE));
-    }
-
+  public PantsExecuteTaskResult invokePants(@NotNull Project currentProject,
+                                            @NotNull Set<String> targetAddresses,
+                                            @NotNull List<String> tasks,
+                                            @NotNull String opTitle) {
+    // TODO: assert that opTitle is one word?
```
wisechengyi: xx

src/com/twitter/intellij/pants/execution/PantsMakeBeforeRun.java
```diff
@@ -173,46 +175,57 @@ public boolean executeTask(
...
     prepareIDE(currentProject);
-    if (targetAddressesToCompile.isEmpty()) {
+
+    List<String> unrecognizedTasks = tasks.stream()
+      .filter(task -> !PantsConstants.SUPPORTED_TASKS.contains(task))
+      .collect(Collectors.toList());
+    if (!unrecognizedTasks.isEmpty()) {
+      String unrecognizedMessage = String.format("Unrecognized tasks specified in Pants invocation: [%s].\n",
```
wisechengyi: this check is probably unnecessary. if the tasks aren't supported, pants will give the corresponding error messages.

src/com/twitter/intellij/pants/execution/PantsMakeBeforeRun.java
```diff
@@ -225,21 +238,17 @@ public boolean executeTask(
...
-    if (useCleanAll) {
-      commandLine.addParameter("clean-all");
-      if (pantsOptions.supportsAsyncCleanAll()) {
-        commandLine.addParameter(PantsConstants.PANTS_CLI_OPTION_ASYNC_CLEAN_ALL);
-      }
+    if (tasks.contains(PantsConstants.PANTS_TASK_CLEAN_ALL) && pantsOptions.supportsAsyncCleanAll()) {
+      commandLine.addParameter(PantsConstants.PANTS_CLI_OPTION_ASYNC_CLEAN_ALL);
     }
-    if (pantsOptions.supportsManifestJar()) {
+    if (tasks.contains(PantsConstants.PANTS_TASK_EXPORT_CLASSPATH) && pantsOptions.supportsManifestJar()) {
```
wisechengyi: the oldest pants version we are validating is 1.2.0, and it has the option, so we can safely remove this check now.

src/com/twitter/intellij/pants/execution/PantsMakeBeforeRun.java
```diff
@@ -250,23 +259,22 @@ public boolean executeTask(
...
       }
     }
 
     /* Goals and targets section. */
-    commandLine.addParameters("export-classpath", "compile");
-    for (String targetAddress : targetAddressesToCompile) {
-      commandLine.addParameter(targetAddress);
-    }
+    commandLine.addParameters(Lists.newArrayList(tasks));
+    commandLine.addParameters(Lists.newArrayList(targetAddresses));
```
wisechengyi: why not just `commandLine.addParameters(tasks)`?

src/com/twitter/intellij/pants/execution/PantsMakeBeforeRun.java
```diff
@@ -285,29 +293,57 @@ public void onTextAvailable(ProcessEvent event, Key outputType) {
...
     final boolean success = process.exitValue() == 0;
-    // Mark project dirty if compile failed.
-    if (!success) {
-      FileChangeTracker.markDirty(currentProject);
-    }
-    else {
-      FileChangeTracker.addManifestJarIntoSnapshot(currentProject);
+    if (tasks.contains(PantsConstants.PANTS_TASK_COMPILE)) {
+      if (success) {
+        // FIXME: is manifest jar always created? what is that? IS it always, if there's a compile?
```
wisechengyi: it is always created if the run succeeds.

src/com/twitter/intellij/pants/execution/PantsMakeBeforeRun.java
```diff
@@ -285,29 +293,57 @@ public void onTextAvailable(ProcessEvent event, Key outputType) {
...
+        FileChangeTracker.addManifestJarIntoSnapshot(currentProject);
+      } else {
+        // Mark project dirty if compile failed.
+        FileChangeTracker.markDirty(currentProject);
+      }
+      // Sync files as generated sources may have changed after Pants compile.
+      PantsUtil.synchronizeFiles();
```
wisechengyi: You are changing the behavior here, now it will only sync if pants run succeeds. We need to sync it regardless.

src/com/twitter/intellij/pants/service/PantsCompileOptionsExecutor.java
```diff
@@ -169,12 +169,12 @@ private String loadProjectStructureFromTargets(
   ) throws IOException, ExecutionException {
     final File outputFile = FileUtil.createTempFile("pants_depmap_run", ".out");
     final GeneralCommandLine command = getCommand(outputFile, statusConsumer);
-    statusConsumer.consume("Resolving dependencies...");
+    statusConsumer.consume("Exporting Pants project into IDEA format...");
```
wisechengyi: it's fine to keep the old one, and it will ease the documentation load.

src/com/twitter/intellij/pants/service/PantsCompileOptionsExecutor.java
```diff
@@ -169,12 +169,12 @@ private String loadProjectStructureFromTargets(
...
     PantsMetrics.markExportStart();
-    final ProcessOutput processOutput = getProcessOutput(command, processAdapter);
+    final ProcessOutput processOutput = getProcessOutput(command);
```
wisechengyi: any reason you are removing the `processAdapter`?

cosmicexplorer: Yes -- it's because the `processAdapter` argument wasn't being used in the `getProcessOutput()` method, so I removed that argument and the argument from all of its callers (just here, I think).

src/com/twitter/intellij/pants/execution/PantsMakeBeforeRun.java
cosmicexplorer: This is outside the `if (success) else` block, so it just runs if one of the tasks is `compile`, which is the same behavior as before since all pants actions call compile as of this commit. I added a blank line before the `// Sync files ...` comment so this is more clear.

src/com/twitter/intellij/pants/execution/PantsMakeBeforeRun.java
cosmicexplorer: Thanks, noted.

src/com/twitter/intellij/pants/execution/PantsMakeBeforeRun.java
cosmicexplorer: Thanks, removed.

src/com/twitter/intellij/pants/execution/PantsMakeBeforeRun.java
cosmicexplorer: removed

src/com/twitter/intellij/pants/service/PantsCompileOptionsExecutor.java
cosmicexplorer: reverted

src/com/twitter/intellij/pants/execution/PantsMakeBeforeRun.java
cosmicexplorer: removed, thanks for the explanation

src/com/twitter/intellij/pants/execution/PantsMakeBeforeRun.java
cosmicexplorer: Thanks for catching that. Changed.

PR #319, opened 2017-11-17
Update to 2017.3

cosmicexplorer:
Pull in version `173.3727.127` of IDEA CE and Ultimate and make some code changes to handle changes in the plugin API.

- update `external-system-test-api.jar`
- update to scala 2.12.4 to match the intellij scala plugin
- change jars used for the scala plugin, and change how jars are selected a little in `3rdparty/intellij/BUILD`
- `PantsUtil.getDefaultJavaSdk()` was being invoked with a directory path in one place, not a pants executable path -- fix that and make a note of it
- errors popped up about not disposing of projects correctly, requiring the change to `PantsUtil.getDefaultJavaSdk()` to accept a `Disposable` and a change to `PantsIntegrationTestCase::tearDown()` to stop errors about jdk table leaks

Small things:
- fixed indentation of `3rdparty/intellij/BUILD` to be consistently 4 spaces
- turned on `strict_deps` for java and scala to help make classpath issues less confusing when finding jars (and I want to keep it on for the same reason in the future, unless I'm missing something)


scripts/prepare-ci-environment.sh
```diff
@@ -69,3 +69,5 @@ append_intellij_jvm_options() {
 
   echo $cmd
 }
+
+export JDK_JARS="$(printf "%s\n" 'sa-jdi.jar' 'tools.jar')"
```
wisechengyi: Since `JDK_JARS` in only used in `scripts/run-tests-ci.sh`, you can probably move it there and do not have to export it.

testFramework/com/twitter/intellij/pants/testFramework/PantsIntegrationTestCase.java
```diff
@@ -495,6 +499,21 @@ public void tearDown() throws Exception {
       killNailgun();
       cleanProjectRoot();
       Messages.setTestDialog(TestDialog.DEFAULT);
+
+      // TODO(dmcclanahan): after updating from 172.4343.14 to 173.3531.6,
+      // intellij's provided test class sometimes yells about a leaky jdk
+      // table. i don't think this indicates any problems, so for now if tests
+      // fail with leaky sdk errors, broaden this to include the leaked sdks
```
wisechengyi: does the comment and following code still apply assuming you fixed it with passing in `getTestRootDisposable()`?

testFramework/com/twitter/intellij/pants/testFramework/PantsIntegrationTestCase.java
```diff
@@ -495,6 +499,21 @@ public void tearDown() throws Exception {
...
+      final Stream<Sdk> leakedJdks = Arrays.stream(ProjectJdkTable.getInstance().getAllJdks())
+        .filter(jdk -> {
+            final String name = jdk.getName();
+            return name.contains("pants") || name.startsWith("python");
```
wisechengyi: are you checking for JDK only or also python interpreter?

tests/com/twitter/intellij/pants/util/PantsUtilTest.java
```diff
@@ -24,19 +27,29 @@ public void testIsPantsProjectFile() {
...
-      @Override
-      public void run() {
-        ProjectJdkTable.getInstance().addJdk(sdkA.get());
-      }
+    final File executable = PantsUtil.findPantsExecutable(getProjectFolder()).get();
+    final Sdk sdkA = getDefaultJavaSdk(executable.getPath()).get();
+    List<Sdk> sdkResults = Arrays.stream(ProjectJdkTable.getInstance().getAllJdks())
+      .filter(jdk -> jdk == sdkA)
+      .collect(Collectors.toList());
+    assertTrue(sdkResults.size() == 1);
```
wisechengyi: `assertEqual` same for below

tests/com/twitter/intellij/pants/util/PantsUtilTest.java
```diff
@@ -24,19 +27,29 @@ public void testIsPantsProjectFile() {
...
+    assertTrue(sdkResults.get(0) == sdkA);
```
wisechengyi: `assertSame` same for below

scripts/prepare-ci-environment.sh
cosmicexplorer: You're right, but I also added a reference to that environment var it in `3rdparty/intellij/BUILD` (I moved the usage to the top of the file to make it more clear) since it was getting duplicated there (the `_jdk_jars` variable, now named `_jdk_jar_paths`).

common/com/twitter/intellij/pants/util/PantsUtil.java
```diff
@@ -801,16 +802,23 @@ public static boolean isGenTarget(@NotNull String address) {
     return addresses.stream().filter(s -> !isGenTarget(s)).collect(Collectors.toSet());
   }
 
-  public static Optional<Sdk> getDefaultJavaSdk(@NotNull final String pantsExecutable) {
-    // If a JDK belongs to this particular `pantsExecutable`, then its name will contain the path to Pants.
+  // NB: `pantsExecutable` should be a path to a pants executable (from
+  // findPantsExecutable()), NOT a directory!
+  // NB: tests should use getTestRootDisposable() here to avoid vfs errors!
```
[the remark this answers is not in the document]
cosmicexplorer: You're right, I pushed a commit making these into a javadoc for that method.

tests/com/twitter/intellij/pants/util/PantsUtilTest.java
cosmicexplorer: Just pushed a commit doing this.

cosmicexplorer: Thanks, pushed a commit which makes the changes.

tests/com/twitter/intellij/pants/util/PantsUtilTest.java
cosmicexplorer: I think using `assertEqual` would do the same check as `assertSame` for each element of the two collections, so we don't need to use it if we're using `assertEqual` to check the identity of each element in the list. Am I missing something?

scripts/run-tests-ci.sh
```diff
@@ -10,19 +10,19 @@ fi
...
+  git pull
+else
+  echo "Using $PANTS_SHA..."
+  git reset --hard $PANTS_SHA
+fi
+popd
+popd
 
-args="test tests:${TEST_SET:-:} $(append_intellij_jvm_options test-junit) ${ADDITIONAL_ARGS:-$@}"
+args="${TASKS:-test} tests:${TEST_SET:-:} $(append_intellij_jvm_options test-junit) ${ADDITIONAL_ARGS:-$@}"
```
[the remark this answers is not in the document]
cosmicexplorer: `TASKS` is either an env var or unset, in which case it expands to just `test`. This was added so I could quickly run different tasks e.g. `./pants classmap tests::` by running `TASKS='classmap' ./scripts/run-tests-ci.sh`.

testFramework/com/twitter/intellij/pants/testFramework/PantsIntegrationTestCase.java
cosmicexplorer: Unfortunately, yes -- the error that is fixed by using `getTestRootDisposable` is not the same as the leaky jdk table error (says "not disposed", while the leaky jdk table error message does not mention disposing of anything). I have not been able to understand why this change in behavior occurs yet.

testFramework/com/twitter/intellij/pants/testFramework/PantsIntegrationTestCase.java
cosmicexplorer: Running the `tests:misc` target with `--test-junit-test=com.twitter.intellij.pants.execution.OSSPantsPythonRunConfigurationIntegrationTest` fails with a leaky jdk table error, and the "leaked" sdk is just named `python:  Python 2.7 (/usr/bin/python)`.

I don't like this, but it seems like the best solution right now. If we expand the number of python imports we do in integration testing I would feel less comfortable, but for now I think this is the only instance. I have made this less of a concern with the most recent commit, only checking for `.startsWith("python")` in `OSSPantsPythonRunConfigurationIntegrationTest#tearDown`.

3rdparty/intellij/BUILD
```diff
@@ -49,94 +46,72 @@ def extract_package_name_version(path):
...
-
-    intellij_jars = []
-
-    # lib
-    for file in os.listdir(os.path.join(base_dir, 'lib')):
-        if file.endswith('.jar'):
-            intellij_jars.append(plugin_jar(os.path.join(base_dir, 'lib', file)))
-
-    return intellij_jars
+def _check_jar_entries(dir, entries):
```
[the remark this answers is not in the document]
cosmicexplorer: Renamed `dir` and `file` vars in 1bf7c5a.

3rdparty/intellij/BUILD
```diff
@@ -49,94 +46,72 @@ def extract_package_name_version(path):
...
+    non_jars = []
+    for entry in entries:
+        full_path = os.path.join(dir, entry)
+        if os.path.isfile(full_path) and full_path.endswith('.jar'): continue
```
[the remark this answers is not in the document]
cosmicexplorer: Fixed in 1bf7c5a.

luisfcofv:
Any update on this PR? I updated to the latest Intellij version and lost support for this plugin.

wisechengyi:
@luisfcofv It's actively being worked on, so hopefully by the end of the week. 

In the meantime, I have an adhoc build for 2017.3 
[pants_1.7.0.1aab2e091a2722ad38f76f037ae945662d326311.zip](https://github.com/pantsbuild/intellij-pants-plugin/files/1528128/pants_1.7.0.1aab2e091a2722ad38f76f037ae945662d326311.zip) (preferences -> plugin -> install from disk), or you may choose to revert to 2017.2.x.

luisfcofv:
Thanks @wisechengyi. I will try out your adhoc build. Thank you!

cosmicexplorer:
@wisechengyi I updated the version info so that it uses version 2017.3 (not the pre-release this time, since it has been fully released), but travis appears to be timing out on shard 2 for the last few commits. I'm not sure why there would be any difference compared to the first?

wisechengyi:
restarted the last ci job. will merge once green

PR #320, opened 2017-11-18
Add assertions for accidental bug fix from #312

cosmicexplorer:
In #312, a change was made to [`CompileAllTargetsInModuleAction.java`](https://github.com/pantsbuild/intellij-pants-plugin/pull/312/files#diff-41cc874d589fdbd635d9ca9f78149ef9) which stopped a bug that affected usage.

To reproduce the error, take a version of the plugin from any commit before b2a024a52a9287d74944400dc4f27a7a613989ba. Open a pants project A, then open a pants project B in a separate window. Click `Build->Compile Module '...'` with a source file from A open in the editor, which should work correctly. Switch to the window for project B, open a source file from B in the editor, then click `Build->Compile Module '...'`. The name of the module will be correct, but the `./pants compile ...` command will build the module from project A (which can be seen in the pants console -- the targets for project A are selected).

#312 accidentally fixed this (see link to diff at the top of this comment) by not mutating the private `module` field in `CompileAllTargetsInModuleAction`, but I didn't realize this modified any behavior until the bug above came to my attention. This adds multiple assertions to `OSSPantsCompileActionsTest#testCompileAllTargetsInModuleAction` to avoid a regression.

### DIFF
--- a/tests/com/twitter/intellij/pants/integration/OSSPantsCompileActionsTest.java
+++ b/tests/com/twitter/intellij/pants/integration/OSSPantsCompileActionsTest.java
@@ -93,12 +93,10 @@ public class OSSPantsCompileActionsTest extends OSSPantsIntegrationTest {
     assertFalse(action.module.isPresent());
 
     VirtualFile vf = PantsUtil.findFileRelativeToBuildRoot(myProject, "testprojects/src/java/org/pantsbuild/testproject/junit/mixed/tests/org/pantsbuild/tmp/tests/AllTests.java").get();
-    Module mockModule = PantsUtil.getModuleForFile(vf, myProject).get();
-    assertEquals(mockModule, mixedModule);
+    assertEquals(mixedModule, PantsUtil.getModuleForFile(vf, myProject).get());
     Set<String> expectedTargetsMixed = Sets.newHashSet(
       "testprojects/src/java/org/pantsbuild/testproject/junit/mixed/tests/org/pantsbuild/tmp/tests:tests");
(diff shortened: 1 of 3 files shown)


tests/com/twitter/intellij/pants/integration/OSSPantsCompileActionsTest.java
```diff
@@ -86,19 +87,50 @@ public void testRebuildAction() throws Throwable {
...
-      .collect(Collectors.toSet());
-    Set<String> expectedTargets = new HashSet<>(Arrays.asList(
+    Module testscopeModule = getModule("testprojects_src_java_org_pantsbuild_testproject_junit_testscope_common_sources");
+    Module mixedModule = getModule("testprojects_src_java_org_pantsbuild_testproject_junit_mixed_tests_org_pantsbuild_tmp_tests_tests");
+    PantsCompileAllTargetsInModuleAction action = new PantsCompileAllTargetsInModuleAction(Optional.empty());
+    assertFalse(action.module.isPresent());
+
+    VirtualFile vf = PantsUtil.findFileRelativeToBuildRoot(myProject, "testprojects/src/java/org/pantsbuild/testproject/junit/mixed/tests/org/pantsbuild/tmp/tests/AllTests.java").get();
+    Module mockModule = PantsUtil.getModuleForFile(vf, myProject).get();
+    assertEquals(mockModule, mixedModule);
```
wisechengyi: nit:

expected value is on the left http://junit.sourceforge.net/javadoc/org/junit/Assert.html#assertEquals(double, double)

same for other assertions

cosmicexplorer: Fixed in 436e514.

PR #321, opened 2017-11-20
Add --no-quiet to Pants export call

wisechengyi:
so Pants will show the error messages if there are any instead of suppressing them. 

There is no effect on how the result is parsed because the json output is written into a file.


wisechengyi:
CI passed but timed out at cache upload. So I am going to merge it.

baroquebobcat:
👍

PR #323, opened 2017-12-09
fix "Unrecognized command line flags on global scope: --async"

cosmicexplorer:
The "Rebuild" action was calling `clean-all`, but it added the `--async` parameter without a thought to the ordering of the command line. This fixes that error by using `--clean-all-async` and adds tests to stop similar regressions (`testRebuildAction` now fails if `--async` is used).


cosmicexplorer:
CI is green!

PR #324, opened 2017-12-20
Pick up .ij.import.rc file for dynamic pantsrc insertion

wisechengyi:
### Problem

Currently all the pants command args from intellij pants plugin are hardcoded from the plugin, so we do not have a way to dynamically tweak per repo, which would be useful if when we want to inject flags for experimental pants features, debug flags, etc.

### Solution
Add a mechanism to recognize <build root>/.ij.import.rc file for dynamic argument changes per repo.


stuhood:
My first thought here is: why not just directly use pants' rc files?

Basically the equivalent of:
```
// At import time.
String importRc = "$buildroot/.ij.import.rc";
if (new File(importRc).isFile()) {
  args.add("--pantsrc-files=" + importRc);
}
```

wisechengyi:
Ah good idea! The only thing is that using pantsrc cannot remove any command line args the plugin passes to pants due to ranked value, but that should not really be an issue.

wisechengyi:
Adjusted.

wisechengyi:
Merging this one since it's a relatively simple change. Like always, if there is any issue, I can always do a follow up patch.

PR #330, opened 2018-02-23
update to 2017.3.4

cosmicexplorer:
JetBrains has released a few versions since we last updated! This PR just updates the version and build number, as well as any checksums. It [passes CI on my fork](https://travis-ci.org/cosmicexplorer/intellij-pants-plugin/builds/344982240), so once it passes CI here I expect it to "just work".


cosmicexplorer:
Just added 18ce89f, which modifies `.travis.yml` to test using pants versions `1.3.0`, `1.2.1`, and `1.4.0.rc1`.

cosmicexplorer:
Reverted the .travis.yml changes as they cause CI to fail. I'll look into injecting `--no-enable-v2-engine` into the invocations of the failing pants versions in a subsequent PR.

PR #331, opened 2018-02-27
Exclude jars imported into the project for IntelliJ Junit/Scala runner

wisechengyi:
### Problem

Currently IntelliJ Junit/Scala runner will bring in all the jars imported into the project to classpath. This is not correct because not all jars imported are needed at runtime, thus causing classpath clashing issues sometimes, e.g. slf4j bindings.

### Solution

Exclude all jars in project except jars on the JDK path, IntelliJ installation paths, and plugin installation path

This patch also disables dynamic classpath for intellij runner since jars in project are excluded from the runner, the command should never be too long.


wisechengyi:
trying to see where I can add some tests..

src/com/twitter/intellij/pants/execution/PantsClasspathRunConfigurationExtension.java
```diff
@@ -63,13 +73,41 @@
...
+      );
+
+      pathsAllowed.add(
+        // At this point we know junit plugin is already enabled.
+        // //Users/yic/.ivy2/pants/com.intellij.junit-plugin/junit-rt/jars/junit-rt-latest.jar ->
+        // /Users/yic/.ivy2/pants/com.intellij.junit-plugin/
+        Optional.ofNullable(PluginManager.getPlugin(PluginId.getId("JUnit")))
+          .map(s -> s.getPath().getParentFile().getParentFile().getParentFile().getAbsolutePath()).get()
+      );
+    }
```
baroquebobcat: I think this would be easier to follow if the code that builds `pathsAllowed` were broken out into a separate method.

src/com/twitter/intellij/pants/execution/PantsClasspathRunConfigurationExtension.java
```diff
@@ -35,18 +45,17 @@
...
   protected static final Logger LOG = Logger.getInstance(PantsClasspathRunConfigurationExtension.class);
   private static final Gson gson = new Gson();
 
   /**
-   * The goal of this function is to find classpath for JUnit runner.
+   * The goal of this function is to find classpath for IntelliJ JUnit/Scala runner.
    * <p/>
-   * There are two ways to do so:
-   * 1. If Pants supports `--export-classpath-manifest-jar-only`, then only the manifest jar will be
+   * Pants needs to support `--export-classpath-manifest-jar-only`, then the manifest jar will be
```
baroquebobcat: Maybe, 
```
This function will fail if the projects' Pants doesn't support `--export-classpath-manifest-jar-only`.
The function assumes that Pants has created a manifest jar that contains all the classpath links for the particular test that's being run.
```

wisechengyi: Done. Thanks!

PR #338, opened 2018-03-28
Force document save before Pants compile action

wisechengyi:
### Problem and Solution

Sometimes when running a test configuration, the plugin misses the change in the project and reuses last compile outputs. Therefore this change forces a save before Pants compile so file tracker will pick up the change before compile action.

### DIFF
--- a/src/com/twitter/intellij/pants/execution/PantsMakeBeforeRun.java
+++ b/src/com/twitter/intellij/pants/execution/PantsMakeBeforeRun.java
@@ -313,6 +313,16 @@ public class PantsMakeBeforeRun extends ExternalSystemBeforeRunTaskProvider {
   public PantsExecuteTaskResult executeCompileTask(@NotNull Project currentProject,
                                                    @NotNull Set<String> targetAddressesToCompile,
                                                    boolean useCleanAll) {
+
+    ApplicationManager.getApplication().invokeAndWait(new Runnable() {
+      @Override
+      public void run() {
+        /* Force cached changes to disk so {@link com.twitter.intellij.pants.file.FileChangeTracker} can mark the project dirty. */
+        FileDocumentManager.getInstance().saveAllDocuments();
+        currentProject.save();
+      }
+    }, ModalityState.NON_MODAL);
+
     // If project has not changed since last Compile, return immediately.
     if (!FileChangeTracker.shouldRecompileThenReset(currentProject, targetAddressesToCompile)) {
       PantsExternalMetricsListenerManager.getInstance().logIsPantsNoopCompile(true);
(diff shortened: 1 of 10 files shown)


src/com/twitter/intellij/pants/execution/PantsMakeBeforeRun.java
```diff
@@ -339,17 +349,14 @@ public void run() {
   }
 
   private void prepareIDE(Project project) {
-    ApplicationManager.getApplication().invokeAndWait(new Runnable() {
+    ApplicationManager.getApplication().invokeLater(new Runnable() {
```
wisechengyi: changed to `invokeLater` because it only contains GUI actions now

src/com/twitter/intellij/pants/execution/PantsMakeBeforeRun.java
```diff
@@ -313,6 +313,16 @@ public void onTextAvailable(ProcessEvent event, Key outputType) {
   public PantsExecuteTaskResult executeCompileTask(@NotNull Project currentProject,
                                                    @NotNull Set<String> targetAddressesToCompile,
                                                    boolean useCleanAll) {
+
+    ApplicationManager.getApplication().invokeAndWait(new Runnable() {
+      @Override
+      public void run() {
+        /* Force cached changes to disk so {@link com.twitter.intellij.pants.file.FileChangeTracker} can mark the project dirty. */
+        FileDocumentManager.getInstance().saveAllDocuments();
+        currentProject.save();
```
wisechengyi: This is the crux of the fix, because the bug was that files were saved after the compile action was called.

testFramework/com/twitter/intellij/pants/testFramework/OSSPantsIntegrationTest.java
```diff
@@ -8,14 +8,21 @@
 import com.intellij.execution.configurations.RunConfiguration;
 import com.intellij.execution.impl.RunManagerImpl;
 import com.intellij.execution.impl.RunnerAndConfigurationSettingsImpl;
+import com.intellij.openapi.editor.Document;
```
wisechengyi: no new functions, only to move some helper functions up from `OSSFileSyncIntegrationTest.java`

wisechengyi:
There is a minor repo inconsistency issue between different pants repo, but this PR is good for review.

src/com/twitter/intellij/pants/execution/PantsMakeBeforeRun.java
baroquebobcat: Does invokeAndWait here have any potential for causing a UI hang, or is this all on a background thread?

wisechengyi: I don't think the code runs on UI thread (inside and outside of `run()`), so no.

wisechengyi:
Thanks for the review guys!

Issue #332, opened 2018-02-28
Build upload broken since 2018


wisechengyi:
https://plugins.jetbrains.com/plugin/7412-pants-support

b71853218b5eddff4d8adcf7a8cb4ff0ea905f39 and 43ec02d80c1a87291c6533ea35997cd82afbba88 are missing

wisechengyi:
filed https://github.com/JetBrains/plugin-repository-rest-client/issues/21

wisechengyi:
also filed https://youtrack.jetbrains.com/issue/IDEA-187960

wisechengyi:
follow up at https://youtrack.jetbrains.com/issue/JT-46456

sugdyzhekov:
Hi! The cause of your problem is a lack of SNI support:
```
* WARNING: disabling hostname validation also disables SNI.
```
https://plugins.jetbrains.com utilizes AWS CloudFront (CDN) to improve service robustness. SNI support is the mandatory requirement of CloudFront.

So, could you try to eliminate `-k` option from curl call?

wisechengyi:
@sugdyzhekov got invalid username/password
```
$ curl -v -i \
> -F userName="..." \
> -F password="..." \
> -F pluginId=7412 \
> -F file=@$(pwd)/pants_1.8.0.b71853218b5eddff4d8adcf7a8cb4ff0ea905f39.zip \
> -F channel=BleedingEdge https://plugins.jetbrains.com/plugin/uploadPlugin
*   Trying 52.84.237.65...
* TCP_NODELAY set
* Connected to plugins.jetbrains.com (52.84.237.65) port 443 (#0)
* TLS 1.2 connection using TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256
* Server certificate: plugins.jetbrains.com
* Server certificate: Amazon
* Server certificate: Amazon Root CA 1
* Server certificate: Starfield Services Root Certificate Authority - G2
> POST /plugin/uploadPlugin HTTP/1.1
> Host: plugins.jetbrains.com
> User-Agent: curl/7.54.0
> Accept: */*
> Content-Length: 263787
> Expect: 100-continue
> Content-Type: multipart/form-data; boundary=------------------------fc36122bf9afa6c7
> 
< HTTP/1.1 100 Continue
HTTP/1.1 100 Continue

< HTTP/1.1 403 
HTTP/1.1 403 
< Content-Type: text/html;charset=utf-8
[...]

PR #340, opened 2018-04-10
Fix auto upload script

wisechengyi:
Addresses #332

Tested manually with 
```
USERNAME='...' PASSWORD='...' TRAVIS_BRANCH=master scripts/deploy/deploy.sh
```


baroquebobcat:
👍

PR #345, opened 2018-04-25
Set up more of the module classpath in the project

illicitonion:
These are all referenced via variables, so should be safe to check in.
Of course, they're a bit version specific, but *shrug* :)


baroquebobcat:
Nice!

PR #346, opened 2018-04-26
Tweak module dependency scope

wisechengyi:
Because all the dependencies here are already available at runtime.


baroquebobcat:
👍

PR #347, opened 2018-04-26
Update python plugin md5

wisechengyi:


wisechengyi:
Merging. Nothing too interesting here.

PR #352, opened 2018-05-30
Make scala plugin an optional dependency

wisechengyi:
This patch allows pants plugin to function without scala plugin. Reason being:
* some folks are only interested in java builds
* whenever there are scala plugin bugs, we will have to be impacted

### DIFF
--- a/common/com/twitter/intellij/pants/util/PantsConstants.java
+++ b/common/com/twitter/intellij/pants/util/PantsConstants.java
@@ -70,6 +70,6 @@ public class PantsConstants {
   public static final String PANTS_TASK_EXPORT_CLASSPATH = "export-classpath";
   public static final String PANTS_TASK_LINT = "lint";
 
-  public static final String SCALA_PLUGIN_ABSTRACT_TEST_RUN_CONFIGURATION_CLASS_NAME =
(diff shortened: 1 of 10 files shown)


dotordogh:
Do you have examples of the bugs you're trying to avoid with this?

common/com/twitter/intellij/pants/util/PantsConstants.java
```diff
@@ -25,6 +25,8 @@
   public static final String PANTS_INI = "pants.ini";
   public static final String PANTS_PEX = "pants.pex";
   public static final String PANTS_LIBRARY_NAME = PANTS_PEX;
+  public static final String ABSTRACT_TEST_RUN_CONFIGURATION_CLASS_NAME =
```
baroquebobcat: Since it's specific to Scala, should this constant have SCALA in it somewhere?

wisechengyi: Good idea!

wisechengyi:
@dotordogh e.g. https://youtrack.jetbrains.com/oauth?state=%2Fissue%2FSCL-13783

wisechengyi:
@baroquebobcat good question on the test. It probably needs a separate run script that does the test without scala plugin on the classpath.

PR #362, opened 2018-07-26
Don't cache lastPath between import wizard runs

baroquebobcat:
- reset lastPath and settings on new import wizard start
- reorder path analysis to allow executable render case to work
- extract path analysis into helper
- use same path analysis for both validation and rendering

### DIFF
--- a/src/com/twitter/intellij/pants/settings/PantsProjectSettingsControl.java
+++ b/src/com/twitter/intellij/pants/settings/PantsProjectSettingsControl.java
@@ -197,7 +197,10 @@ public class PantsProjectSettingsControl extends AbstractExternalProjectSettings
         });
         break;
       default:
-        throw new ConfigurationException("unexpected state: " + pathFileType);
+        UIUtil.invokeLaterIfNeeded((Runnable) () -> {
+          Messages.showErrorDialog(getProject(), "Unexpected project file state: " + pathFileType, "Pants Failure");
+          Messages.createMessageDialogRemover(getProject()).run();
+        });
     }
   }
 
@@ -227,7 +230,7 @@ public class PantsProjectSettingsControl extends AbstractExternalProjectSettings
         }
         break;
       default:
-        throw new ConfigurationException("unexpected state: " + pathFileType);
+        throw new ConfigurationException("Unexpected project file state: " + pathFileType);
     }
     if (!errors.isEmpty()) {
       String errorMessage = String.join("\n", errors);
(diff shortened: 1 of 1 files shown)


src/com/twitter/intellij/pants/settings/PantsProjectSettingsControl.java
```diff
@@ -112,65 +120,84 @@ protected boolean isExtraSettingModified() {
...
                 Messages.createMessageDialogRemover(getProject()).run();
               });
+            }
           }
-        }
-      });
+        });
+        break;
+      default:
+        throw new ConfigurationException("unexpected state: " + pathFileType);
```
wisechengyi: The compile error suggests this function cannot throw `ConfigurationException`, so using `UIUtil.invokeLaterIfNeeded` like above might be more appropriate.

baroquebobcat: last minute adjustment fail 🤦‍♂️

baroquebobcat: Does `UIUtil.invokeLaterIfNeeded` allow the exception to bubble up somewhere visible?

wisechengyi: Yes the example above has `Pants Failure: ...`, which shows up as 
screen shot 2018-07-26 at 1 47 50 pm

You can test it by changing pants script to return false
```
$ cat pants
false
```

src/com/twitter/intellij/pants/settings/PantsProjectSettingsControl.java
```diff
@@ -112,65 +120,87 @@ protected boolean isExtraSettingModified() {
...
                 Messages.createMessageDialogRemover(getProject()).run();
               });
+            }
           }
-        }
-      });
+        });
+        break;
+      default:
+        UIUtil.invokeLaterIfNeeded((Runnable) () -> {
```
wisechengyi: Hmm selecting a directory would trigger this..so maybe noop, which is the status quo?

baroquebobcat: hm. I don't think so. It should hit https://github.com/pantsbuild/intellij-pants-plugin/pull/362/files/655250383b1e924edb3c14e9d144a69aa9f1c5b6#diff-42c61d4fa92b800ccf4d203aca9fc5feR151

wisechengyi: screen shot 2018-07-26 at 3 10 55 pm

baroquebobcat: nice

Issue #333, opened 2018-03-09
A "hack to trick BuildProcessClasspathManager" won't work in 2018.2+


trespasserw:
The `PantsInitComponentImpl#initComponent` method calls `IdeaPluginDescriptorImpl#setPath` (at [PantsInitComponentImpl.java:50](https://github.com/pantsbuild/intellij-pants-plugin/blob/master/src/com/twitter/intellij/pants/components/impl/PantsInitComponentImpl.java#L50)). In IDEA 2018.2, the call will have no effect (the method is deprecated and scheduled for removal).

If I correctly understood the goal of the hack, you can achieve the same by implementing `BuildProcessParametersProvider` extension. Please let me know if you need any help with that.

wisechengyi:
Thanks for the heads up, @trespasserw! We plan to get to the deprecation issue soon.

PR #363, opened 2018-07-27
IntelliJ 2018.2 Compatibility

### LINKED ISSUE #360: intellij builds above 182 not supported
The latest version (2018.2) will no longer load the plugin.

I see a few issues (could be more):
1)  plugin.xml says not to: `<idea-version since-build="181.0" until-build="181.*"/>`
2) `PantsSetttings`  uses `StoragePathMacros.PROJECT_CONFIG_DIR` which no longer exists

wisechengyi:
Resolves #360 #333

### Major changes
* `StoragePathMacros` no longer exists in the core, so `PantsSettings` will live in `misc.xml`
* `compileServer.plugin` is no longer and should not be used, only leaving `PantsBuildProcessParametersProvider` to throw an error in case user accidentally ran into it.

### DIFF
--- a/resources/META-INF/plugin.xml
+++ b/resources/META-INF/plugin.xml
@@ -15,7 +15,7 @@
   <version>1.9.2</version>
   <vendor>Twitter, Inc.</vendor>
 
-  <idea-version since-build="181.0" until-build="181.*"/>
+  <idea-version since-build="182.0" until-build="182.*"/>
 
   <!--Add gradle as a dep because of Pants runner requires it.-->
   <depends>org.jetbrains.plugins.gradle</depends>
--- a/resources/META-INF/plugin.xml
+++ b/resources/META-INF/plugin.xml
@@ -119,7 +119,6 @@
                      groupName="Pants" enabledByDefault="true"
                      implementationClass="com.twitter.intellij.pants.inspection.PythonPluginInspection"/>
 
-    <!--<compileServer.plugin classpath="intellij-pants-plugin-publish.jar"/>-->
     <buildProcess.parametersProvider implementation="com.twitter.intellij.pants.compiler.PantsBuildProcessParametersProvider"/>
 
     <projectConfigurable groupId="build.tools" groupWeight="110" id="reference.settingsdialog.project.pants"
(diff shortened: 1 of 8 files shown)


resources/META-INF/plugin.xml
```diff
@@ -119,7 +119,6 @@
                      groupName="Pants" enabledByDefault="true"
                      implementationClass="com.twitter.intellij.pants.inspection.PythonPluginInspection"/>
 
-    <compileServer.plugin classpath="intellij-pants-plugin-publish.jar"/>
```
wisechengyi: We used to just throw an error if we reached external build process by accident, and now that's been moved to `PantsBuildProcessParametersProvider`, so no need for this anymore.

src/com/twitter/intellij/pants/components/impl/PantsInitComponentImpl.java
```diff
@@ -43,13 +37,6 @@ public void initComponent() {
     final String key = PantsConstants.SYSTEM_ID.getId() + ExternalSystemConstants.USE_IN_PROCESS_COMMUNICATION_REGISTRY_KEY_SUFFIX;
     Registry.get(key).setValue(true);
 
-    // hack to trick BuildProcessClasspathManager
```
wisechengyi: Not needed since we don't use external builder anymore.

src/com/twitter/intellij/pants/settings/PantsSettings.java
```diff
@@ -23,11 +20,7 @@
 import java.util.Set;
 
 @State(
-  name = "PantsSettings",
-  storages = {
-    @Storage(file = StoragePathMacros.PROJECT_FILE),
-    @Storage(file = StoragePathMacros.PROJECT_CONFIG_DIR + "/pants.xml", scheme = StorageScheme.DIRECTORY_BASED)
-  }
```
wisechengyi: this now lives in `misc.xml`

MasterDDT:
Hi @wisechengyi, is this gonna be released soon?

wisechengyi:
Yes, should be this week. You can download the master build already at https://plugins.jetbrains.com/plugin/7412-pants-support/update/48241 then install manually.

PR #367, opened 2018-07-30
Bump scala plugin sha

wisechengyi:


wisechengyi:
Going to merge. Nothing too interesting here.

PR #380, opened 2018-11-27
update plugin to support 2018.3

mrkb80:
Gave a quick pass at supporting 2018.3.  Have a look and let me know if I can do anything else.

### DIFF
--- a/resources/META-INF/plugin.xml
+++ b/resources/META-INF/plugin.xml
@@ -15,7 +15,7 @@
   <version>1.10.1</version>
   <vendor>Twitter, Inc.</vendor>
 
-  <idea-version since-build="182.0" until-build="182.*"/>
+  <idea-version since-build="182.0" until-build="183.*"/>
 
   <!--Add gradle as a dep because of Pants runner requires it.-->
   <depends>org.jetbrains.plugins.gradle</depends>
--- a/src/com/twitter/intellij/pants/PantsManager.java
+++ b/src/com/twitter/intellij/pants/PantsManager.java
@@ -40,11 +40,9 @@ import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 
 import javax.swing.Icon;
-import java.net.URL;
-import java.util.List;
 import java.util.Optional;
 
-public class PantsManager implements
+public class  PantsManager implements
                           ExternalSystemConfigurableAware,
                           ExternalSystemUiAware,
                           StartupActivity,
@@ -181,10 +179,6 @@ public class PantsManager implements
     parameters.getVMParametersList().addProperty(
       ExternalSystemConstants.EXTERNAL_SYSTEM_ID_KEY, PantsConstants.SYSTEM_ID.getId()
     );
-  }
-
-  @Override
-  public void enhanceLocalProcessing(@NotNull List<URL> urls) {
   }
 
   @Override
--- a/src/com/twitter/intellij/pants/compiler/PantsBuildProcessParametersProvider.java
+++ b/src/com/twitter/intellij/pants/compiler/PantsBuildProcessParametersProvider.java
@@ -7,9 +7,9 @@ import com.intellij.compiler.server.BuildProcessParametersProvider;
 import com.intellij.openapi.project.Project;
 import com.twitter.intellij.pants.util.PantsConstants;
 import com.twitter.intellij.pants.util.PantsUtil;
-import org.fest.util.Lists;
 import org.jetbrains.annotations.NotNull;
 
+import java.util.ArrayList;
 import java.util.List;
 
 public class PantsBuildProcessParametersProvider extends BuildProcessParametersProvider {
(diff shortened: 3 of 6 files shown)


resources/META-INF/plugin.xml
```diff
@@ -15,7 +15,7 @@
   <version>1.10.1</version>
   <vendor>Twitter, Inc.</vendor>
 
-  <idea-version since-build="182.0" until-build="182.*"/>
+  <idea-version since-build="182.0" until-build="183.*"/>
```
wisechengyi: If it is not backwards compatible with 182.*, we may have to do the following

```suggestion
  <idea-version since-build="183.0" until-build="183.*"/>
```

src/com/twitter/intellij/pants/PantsManager.java
```diff
@@ -183,10 +181,6 @@ public void enhanceRemoteProcessing(@NotNull SimpleJavaParameters parameters) th
     );
   }
 
-  @Override
-  public void enhanceLocalProcessing(@NotNull List<URL> urls) {
-  }
-
   @Override
   public void runActivity(@NotNull Project project) {
   }
```
wisechengyi: Change in this file seems unrelated. We can land a separate patch for beautification.

src/com/twitter/intellij/pants/compiler/PantsBuildProcessParametersProvider.java
```diff
@@ -27,6 +27,6 @@ public PantsBuildProcessParametersProvider(Project project) {
     if (PantsUtil.isPantsProject(myProject)) {
       throw new RuntimeException(PantsConstants.EXTERNAL_BUILDER_ERROR);
     }
-    return Lists.newArrayList();
```
wisechengyi: Ah this was meant to use the guava `Lists`

tests/com/twitter/intellij/pants/settings/PantsProjectSettingsTest.java
```diff
@@ -7,9 +7,10 @@
 import com.intellij.ui.CheckBoxList;
 import com.intellij.util.containers.ContainerUtil;
 import com.twitter.intellij.pants.testFramework.OSSPantsImportIntegrationTest;
-import org.fest.util.Sets;
```
wisechengyi: same goes for guava sets for ease of use.

resources/META-INF/plugin.xml
mrkb80: I think we will need to change this since some APIs changed.

src/com/twitter/intellij/pants/compiler/PantsBuildProcessParametersProvider.java
mrkb80: I figured, but rather than add guava, I just used the base java methods.  We could also add guava (big fan!).

tests/com/twitter/intellij/pants/settings/PantsProjectSettingsTest.java
mrkb80: :)

src/com/twitter/intellij/pants/compiler/PantsBuildProcessParametersProvider.java
wisechengyi: guava is already used by intellij core, so there's no additional cost to use it.

src/com/twitter/intellij/pants/PantsManager.java
wisechengyi: sounds good.

wisechengyi:
JFYI still need to change the version that CI tests against. Example of 2018.2 upgrade: https://github.com/pantsbuild/intellij-pants-plugin/commit/ba8b92e159d2711ac53e0235dce63b0bb38d1a49#diff-3b968e6f002f17073e558a7bf4c32e71

src/com/twitter/intellij/pants/compiler/PantsBuildProcessParametersProvider.java
mrkb80: let's use guava then.

wisechengyi:
Also just a tip: for iteration you can comment out https://github.com/pantsbuild/intellij-pants-plugin/blob/master/scripts/setup-ci-environment.sh#L12, then update the SHAs later.

mrkb80:
> Also just a tip: for iteration you can comment out https://github.com/pantsbuild/intellij-pants-plugin/blob/master/scripts/setup-ci-environment.sh#L12, then update the SHAs later.

ah, that will make this go a bit faster.  Let me get this fixed up now.

mrkb80:
@wisechengyi any idea why this test is failing (there seems to be a TODO indicating a failure is expected?):
`There was 1 failure:
                     1) testScalaRunnerMetrics(com.twitter.intellij.pants.extension.PantsExternalMetricsListenerExtensionTest)
                     java.lang.ClassCastException: com.intellij.execution.compound.CompoundRunConfigurationOptions cannot be cast to com.intellij.execution.configurations.RunConfigurationOptions`

wisechengyi:
will take a look tomorrow. sounds like a breaking api change tho.

wisechengyi:
will check which test is hanging for more than 10 min.

wisechengyi:
I can repro the hang locally, but unable to determine which test is hanging yet. If you could kindly help identify the hanging test(s), that'll be helpful @mrkb80.

It is also a possibility that no particular tests are hanging, but a series of tests left things behind together that caused the hang.

wisechengyi:
will merge once the last commit is green.

mrkb80:
thanks!

PR #381, opened 2018-11-28
populate the environment of the pants subprocess with a subshell's env

cosmicexplorer:
Some users internally are seeing this error message nondeterministically when attempting to run a test:
```
pants.bin.pants_loader.InvalidLocaleError: System preferred encoding is `US-ASCII`, but `UTF-8` is required.
Check and set the LC_* and LANG environment settings. Example:
  LC_ALL=en_US.UTF-8
  LANG=en_US.UTF-8
To bypass this error, please file an issue and then set:
  PANTS_IGNORE_UNRECOGNIZED_ENCODING=1
```

This occurs even though the user has set these variables in their shell. This PR uses `EnvironmentUtil.ShellEnvReader` to pull variables from an invocation of a subshell (the implementation in intellij invokes a python script).

It's not clear when and why these variables are changing, and I haven't yet figured out how to test that using the shell environment as in this PR actually resolves this issue, as it is occurring nondeterministically.


tests/com/twitter/intellij/pants/util/PantsUtilTest.java
```diff
@@ -57,6 +66,38 @@ public void testFindJdk() {
...
+  }
+
+  public void testSubprocessWithShellEnvSucceeds() throws ExecutionException {
+    final GeneralCommandLine command = PantsUtil.defaultCommandLine(getPantsExecutablePath());
+    command.addParameter("options");
+
+    final ProcessOutput processOutput = PantsUtil.getCmdOutput(command, null);
+    assertTrue(processOutput.checkSuccess(LOG));
+  }
+
```
wisechengyi: The tests look good and should stay, but they are not specifically testing the case where intellij was launched with the bad `LANG` or `LC_ALL` env values and then the pants invocation still works, in case that was what you intended to test.

common/com/twitter/intellij/pants/util/PantsUtil.java
```diff
@@ -353,6 +354,15 @@ public static GeneralCommandLine defaultCommandLine(@NotNull File pantsExecutabl
       pantsExecutable.getAbsolutePath()
     );
     commandLine.setExePath(pantsExecutablePath);
+
+    // Seed the subprocess invocation with the environment of a shell (TODO: which shell?).
+    try {
+      Map<String, String> shellEnv = new EnvironmentUtil.ShellEnvReader().readShellEnv();
```
wisechengyi: It is reading env vars from a new subshell which reloads whatever it is in `~/.bash_profile`, this would be fine. Can you kindly make sure this is indeed the behavior here like mentioned in the comment?

tests/com/twitter/intellij/pants/util/PantsUtilTest.java
cosmicexplorer: That is exactly what I intended to test! I guess an integration test could do this exactly — wasn’t thinking too hard about it before.

wisechengyi: Sounds good. This is fine as is, but feel free to add more tests.

tests/com/twitter/intellij/pants/util/PantsUtilTest.java
```diff
@@ -57,6 +66,38 @@ public void testFindJdk() {
...
+  public void testNonUtf8LocaleSubprocessFails() throws ExecutionException {
+    final GeneralCommandLine command = PantsUtil.defaultCommandLine(getPantsExecutablePath());
+    command.addParameter("options");
+
+    final Map<String, String> localeOverrides = new HashMap<>();
+    final String badLocale = "en_US.US-ASCII";
+    localeOverrides.put("LANG", badLocale);
+    localeOverrides.put("LC_ALL", badLocale);
```
wisechengyi: nit: you can probably use some of the guava functions to simply this to oneliner.

tests/com/twitter/intellij/pants/util/PantsUtilTest.java
```diff
@@ -57,6 +66,38 @@ public void testFindJdk() {
     assertEquals(twoEntriesSameSdk, getSameJdks(sdkA));
   }
 
+  protected String getPantsExecutablePath() {
+    return PantsUtil.findPantsExecutable(getParentPath()).get().getPath();
+  }
+
+  public void testNonUtf8LocaleSubprocessFails() throws ExecutionException {
+    final GeneralCommandLine command = PantsUtil.defaultCommandLine(getPantsExecutablePath());
+    command.addParameter("options");
```
wisechengyi: nit: move line 74, 75 below line 80.

tests/com/twitter/intellij/pants/util/PantsUtilTest.java
```diff
@@ -4,19 +4,28 @@
 package com.twitter.intellij.pants.util;
 
 import com.google.common.collect.Lists;
+import com.intellij.execution.ExecutionException;
+import com.intellij.execution.configurations.GeneralCommandLine;
+import com.intellij.execution.process.CapturingProcessHandler;
```
wisechengyi: unused.

PR #390, opened 2019-01-24
Add import success notification

wisechengyi:
7a60398f83b7c98cba8e6a1a9891082de7ff1aa7 disabled focusing on the imported dir, but users may be confused at first after the project finishes importing. This makes the import success more explicit.

### DIFF
--- a/src/com/twitter/intellij/pants/service/project/PantsSystemProjectResolver.java
+++ b/src/com/twitter/intellij/pants/service/project/PantsSystemProjectResolver.java
@@ -94,7 +94,6 @@ public class PantsSystemProjectResolver implements ExternalSystemProjectResolver
 
   private void sendSuccessMessage(@NotNull ExternalSystemTaskId id) {
     ApplicationManager.getApplication().invokeLater(() -> {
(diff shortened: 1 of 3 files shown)


src/com/twitter/intellij/pants/service/project/PantsSystemProjectResolver.java
```diff
@@ -82,11 +85,26 @@
...
     task2executor.remove(id);
     return projectDataNode;
   }
 
+  private void sendSuccessMessage(@NotNull ExternalSystemTaskId id) {
+    ApplicationManager.getApplication().invokeLater(() -> {
+      Notification myNotification = new Notification(
+        PantsConstants.PANTS,
+        "Success",
+        "Pants project import/refresh succeeded.",
```
baroquebobcat: Should this string live in https://github.com/pantsbuild/intellij-pants-plugin/blob/master/common/com/twitter/intellij/pants/PantsBundle.properties?

wisechengyi: ah yes. will adjust.

PR #401, opened 2019-04-15
[Incremental import] Auto target alias expansion

wisechengyi:
### Problem

Previously with incremental import, target aliases `target` and `alias` will get counted as one level, causing user having to expand more levels.

### Solution

This patch adds the mechanism that if a target alias is included in the query, it automatically adds its dependencies into the final results.


wisechengyi:
failure is known and unrelated. will address in a separate batch

PR #404, opened 2019-06-06
If there are a large number of target specs, pass them to pants via a file to avoid exceeding the OS command line length limit.

sundresh:
If there are a large number of target specs, pass them to pants via a file to avoid exceeding the OS command line length limit.

Specifically uses a threshold of 100 right now, which is directly in the code as opposed to being defined in a constants file. This seems to be consistent with the rest of the plugin code, which is littered with strings, although there is at least one shared constant `PantsCompileOptionsExecutor.PROJECT_NAME_LIMIT`. In practice, the threshold of 100 should not make a difference to the rest of the plugin code.

Does not currently include any unit tests. In principle it seems like it should be possible to unit test `PantsCompileOptionsExecutor.getPantsExportCommand()`.

### DIFF
--- a/src/com/twitter/intellij/pants/service/PantsCompileOptionsExecutor.java
+++ b/src/com/twitter/intellij/pants/service/PantsCompileOptionsExecutor.java
@@ -28,6 +28,7 @@ import org.jetbrains.annotations.Nullable;
 import org.jetbrains.annotations.TestOnly;
 
 import java.io.File;
+import java.io.FileWriter;
 import java.io.IOException;
 import java.util.Collections;
 import java.util.List;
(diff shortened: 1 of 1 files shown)


stuhood:
@sundresh : pants has a `--target-spec-files=$file` option, which specifies a file containing target specs:
```
$ ./pants help
...
--target-spec-file=<str> (--target-spec-file=<str>) ..., --target-spec-file="[<str>, <str>, ...]", --target-spec-file="+[<str>, <str>, ...]" (default: [])
    Read additional specs from this file, one per line
```

sundresh:
> @sundresh : pants has a `--target-spec-file=$file` option, which specifies a file containing target specs:
> 
> ```
> $ ./pants help
> ...
> --target-spec-file=<str> (--target-spec-file=<str>) ..., --target-spec-file="[<str>, <str>, ...]", --target-spec-file="+[<str>, <str>, ...]" (default: [])
>     Read additional specs from this file, one per line
> ```
> 
> I think that you should use that instead. Multiple calls to `export` will definitely not be consistent with regard to the classpath.

Thanks for bringing that to my attention, I should've checked for something like that. I've updated this PR to use `--target-spec-file` if there are > 100 target specs.

Moving the missing `equals()`, `hashCode(`) and `toString()` methods to a separate PR, since they are no longer necessary for this change but may still be nice to have.

PR #405, opened 2019-06-18
Move --target-spec-file before goals and use target spec file only

wisechengyi:
To fix:
```
pants --pantsrc-files=/Users/srohankar/devel_workspace/source/.ij.import.rc --no-quiet export --formatted --export-libraries-sources --export-libraries-javadocs --target-spec-file=/private/var/folders/h7/zbp8j4zj6jd8mpvcvbf4cbhw0000gn/T/pants_target_specs1.in --export-output-file=/private/var/folders/h7/zbp8j4zj6jd8mpvcvbf4cbhw0000gn/T/pants_depmap_run2.out
Exit code: 1
...
Exception message: Unrecognized command line flags on scope 'export': --target-spec-file. Suggestions:
--target-spec-file: [--target-spec-file]
```


wisechengyi:
cc @sundresh

PR #408, opened 2019-07-05
update to work with 2019.2 EAP2

DuncanHills:
there are still some problems with settings serialization, but this is a start.

### DIFF
--- a/common/common.iml
+++ b/common/common.iml
@@ -7,5 +7,6 @@
     </content>
     <orderEntry type="inheritedJdk" />
     <orderEntry type="sourceFolder" forTests="false" />
+    <orderEntry type="library" name="scala-sdk-2.12.8" level="application" />
   </component>
 </module>
--- a/intellij-pants-plugin.iml
+++ b/intellij-pants-plugin.iml
@@ -35,6 +35,19 @@
         <JAVADOC />
         <SOURCES>
           <root url="jar://$APPLICATION_PLUGINS_DIR$/Scala/lib/scalaUltimate.jar!/com/intellij/scala/play/projectTemplate/sbtintegr/skeleton260/app_java" />
+          <root url="file://$MODULE_DIR$/../intellij-scala/bsp/src" />
+          <root url="file://$MODULE_DIR$/../intellij-scala/scala/codeInsight/src" />
+          <root url="file://$MODULE_DIR$/../intellij-scala/scala/compiler-jps/src" />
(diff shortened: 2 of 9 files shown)


common/common.iml
```diff
@@ -7,5 +7,6 @@
     </content>
     <orderEntry type="inheritedJdk" />
     <orderEntry type="sourceFolder" forTests="false" />
+    <orderEntry type="library" name="scala-sdk-2.12.8" level="application" />
```
wisechengyi: This change may not be needed, but if it is, I'd imagine this would need to be 'provided' or 'compile-only' scope. Otherwise scala sdk would leak into runtime, conflicting with scala plugin.

intellij-pants-plugin.iml
```diff
@@ -35,6 +35,19 @@
...
+          <root url="file://$MODULE_DIR$/../intellij-scala/scala/compiler-shared/src" />
+          <root url="file://$MODULE_DIR$/../intellij-scala/scala/conversion/src" />
+          <root url="file://$MODULE_DIR$/../intellij-scala/scala/integration/devKit/src" />
+          <root url="file://$MODULE_DIR$/../intellij-scala/scala/integration/intellilang/src" />
+          <root url="file://$MODULE_DIR$/../intellij-scala/scala/nailgun/src" />
+          <root url="file://$MODULE_DIR$/../intellij-scala/scala/runners/src" />
+          <root url="file://$MODULE_DIR$/../intellij-scala/scala/scala-api/src" />
+          <root url="file://$MODULE_DIR$/../intellij-scala/scala/scala-impl/src" />
+          <root url="file://$MODULE_DIR$/../intellij-scala/scala/scala-impl/test" />
+          <root url="file://$MODULE_DIR$/../intellij-scala/scala/worksheet/src" />
```
wisechengyi: might be easier not to commit this file, as this seems local specific content.

resources/META-INF/plugin.xml
```diff
@@ -12,10 +12,10 @@
       </ul>
       ]]>
   </change-notes>
-  <version>1.12.1</version>
+  <version>1.12.2</version>
```
wisechengyi: let's bump it to 1.13.0 as we typically do for major release upgrades.

ggetv:
do you know when we can have a new release compatible with 2019.2?

wisechengyi:
close in favor of #409

PR #409, opened 2019-07-27
Update intellij-pants-plugin to work with IntelliJ IDEA CE 2019.2

sundresh:
* Some APIs have changed in IntelliJ IDEA 2019.2; @DuncanHills updated the plugin to use the new form of these APIs. Due to these API changes, this update to the plugin is not compatible with IntelliJ 2019.1; I've updated `plugin.xml` accordingly.
* IntelliJ 2019.2 also requires certain actions to either be run or not run on the event dispatcher thread. I fixed the cases I could find where we were previously calling these actions on the wrong thread.


sundresh:
@wisechengyi I updated `IJ_VERSION` and `IJ_BUILD_NUMBER` to try to fix CI build.

scripts/prepare-ci-environment.sh
```diff
@@ -10,8 +10,8 @@ fi
 # Python plugin for Community and Ultimate Edition
 
 export CWD=$(pwd)
-export IJ_VERSION="191.6183.62"
-export IJ_BUILD_NUMBER="191.6183.62"
+export IJ_VERSION="192.5728.98"
```
wisechengyi: This is the build number if it's EAP, and the release name if it's released. so `2019.2` in this case.

wisechengyi:
```suggestion
export IJ_VERSION="2019.2"
```

wisechengyi:
After bumping to 2019.2, some of the compilation issues are legit, because jetbrains moved their API around.

scripts/prepare-ci-environment.sh
sundresh: I've applied this suggestion and added a comment. This was causing `scripts/setup-ci-environment.sh` to fail with exit code 8 when attempting to download IntelliJ.

sundresh:
> After bumping to 2019.2, some of the compilation issues are legit, because jetbrains moved their API around.

Hopefully there aren't any, since I've been locally building the plugin with 2019.2.

wisechengyi:
Would need to make sure the sdk is also 2019.2 as well. I do see the IntelliJ indicating certain symbols as unresolved. 

Screen Shot 2019-07-30 at 12 20 03 PM

sundresh:
> Would need to make sure the sdk is also 2019.2 as well. I do see the IntelliJ indicating certain symbols as unresolved.
> 
> Screen Shot 2019-07-30 at 12 20 03 PM

I don't see that problem in IntelliJ, but I can reproduce the Pants test failure. So I must have updated the SDK version within some aspect of the IntelliJ project that doesn't get checked in. How do I update the IntelliJ SDK version used in the repo?

scripts/prepare-ci-environment.sh
```diff
@@ -10,8 +10,10 @@ fi
 # Python plugin for Community and Ultimate Edition
 
 export CWD=$(pwd)
-export IJ_VERSION="191.6183.62"
-export IJ_BUILD_NUMBER="191.6183.62"
+# Normally, IJ_VERSION is of the form YEAR.x[.y[.z]]
+# But for EAPs, set IJ_VERSION to the same as IJ_BUILD_NUMBER
```
Sohamr: nit: we can remove these comments as this is no longer for EAP version.

wisechengyi: It is still needed actually to remind people how this works, e.g. when 2019.3 comes around.

sundresh:
I discovered that I was still using a 2019.2 EAP version of the IntelliJ SDK in my IntelliJ project for this plugin, which might explain the inconsistency with CI. I first tried updating my existing project to the latest IntelliJ SDK version, but builds kept working. So I tried deleting my project, running `git clean -fdx` in the plugin repo, and then re-opening the project and following the instructions in `CONTRIBUTING.md`. Now I have a _different_ set of errors:
```
/Users/ssundresh/workspace/intellij-pants-plugin/testFramework/com/twitter/intellij/pants/testFramework/performance/PantsPerformanceBenchmark.scala
Error:(15, 8) object PantsTargetAddress is not a member of package com.twitter.intellij.pants.model
import com.twitter.intellij.pants.model.PantsTargetAddress
Error:(17, 8) object PantsUtil is not a member of package com.twitter.intellij.pants.util
import com.twitter.intellij.pants.util.PantsUtil
Error:(55, 19) not found: value PantsTargetAddress
    val address = PantsTargetAddress.fromString(path)
Error:(56, 21) not found: value PantsUtil
    val buildRoot = PantsUtil.findBuildRoot(new File(address.getPath))
```
Interestingly, IntelliJ doesn't flag those imports as errors in the IDE inspections, and correctly allows me to jump to their definitions. So something is going wrong at build time only...

wisechengyi:
Now it fails at:
```
                     39) testHello(com.twitter.intellij.pants.integration.OSSPantsJavaExamplesIntegrationTest)
2343                     junit.framework.AssertionFailedError: org.jetbrains.plugins.gradle is not enabled!
2344                     	at junit.framework.Assert.fail(Assert.java:57)
2345                     	at junit.framework.Assert.assertTrue(Assert.java:22)
```
would need to investigate why gradle plugin isn't enabled in this case. Possibly upstream has changed test frameworks' behavior.

DuncanHills:
looks like this is getting very close! ᕕ( ᐛ )ᕗ

tests/com/twitter/intellij/pants/components/impl/PantsProjectCacheTest.java
```diff
@@ -43,7 +43,15 @@ public void run() {
   @Override
   protected void tearDown() throws Exception {
     PantsProjectCacheImpl.getInstance(getProject()).projectClosed();
-    super.tearDown();
+    try {
+      super.tearDown();
+    }
+    catch (Exception e) {
+      // TODO: something in the test framework isn't disposed.
```
DuncanHills: It looks like this is `PsiTestUtil.addSourceRoot(getModule(), VfsUtil.createDirectoryIfMissing(root, "bar"));`. This directory is likely being created and thus needs to be destroyed as well. Eating these exceptions will probably mask actual resource leaks.

wisechengyi: You are right, although I haven't figured out how to do that yet.

Since this isn't critical since it's an issue from test framework, I've decided to unblock people from getting updates for now.

wisechengyi: Separate PR up: https://github.com/pantsbuild/intellij-pants-plugin/pull/412

PR #413, opened 2019-08-08
Fix upload with token

wisechengyi:


wisechengyi:
Merging this since this has no change on the plugin itself.

PR #416, opened 2019-09-20
Integrate Azure Pipelines

wisechengyi:
Haven't figured out the publishing story yet, so will keep travis around for a bit more, but can only use it for publishing rather than test once this lands.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -18,20 +18,12 @@ jobs:
 
     strategy:
       matrix:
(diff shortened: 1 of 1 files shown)


wisechengyi:
Azure pipeline result visible at my fork: https://dev.azure.com/wisechengyi0295/wisechengyi/_build/results?buildId=41

Need pantsbuild admin to enable azure pipeline for this repo.

PR #424, opened 2019-10-08
Throw explicit ExternalSystemException upon Pants run exit non-zero

wisechengyi:


src/com/twitter/intellij/pants/service/task/PantsTaskManager.java
```diff
@@ -68,6 +68,7 @@ public void executeTasks(
     listener.onTaskOutput(id, commandLine.getCommandLineString(PantsConstants.PANTS), true);
     try {
       final Process process = commandLine.createProcess();
+      List<String> textOutputs = Lists.newArrayList();
```
Sohamr: Curious: Why not use List instead of String?

wisechengyi: hmm no particular difference in this case, but possibly easier to read, manipulate, and debug.

PR #427, opened 2019-11-13
Replace Sdk with a Jdk reference as the project data

marek1840:
This closes #426 by replacing the Sdk with a Jdk reference data class (with java home and jsk name). The data class, unlike the Sdk, doesn't contain any primitives unsupported by the IntelliJ serializer.

PR #430, opened 2019-11-14
Add tests for JdkRef

marek1840:
This is a follow up to #427


wisechengyi:
Thanks, @marek1840! Looks like the error is legit, you can try to repro locally via
```
./scripts/run-tests-ci.sh --test-junit-test=com.twitter.intellij.pants.model.JdkRefTest
```

marek1840:
> Thanks, @marek1840! Looks like the error is legit (...)

Yup, turns out that:
1. I was not disposing the PantsConsole after the test
2. I missed one place where I needed to pass the Disposable when creating a JDK

PR #438, opened 2019-11-22
Add toggles for importing source deps as jars and using the intellij compiler

blorente:
This PR contains 2 changes:

- Adding a toggle to the project import dialogue for importing source dependencies as jars. The value of this option is also read from the project description exported by `./pants idea-plugin`, and is passed down to the calls of `export` made by pants.
- Adding a toggle to the project import dialogue to specify whether we want to use the IntelliJ compiler (as opposed to pants) to compile code. The value of this option gets stored in `PantsProjectSettings`, but it currently not used anywhere.


wisechengyi:
There might be some lifecycle issue not related to your change, but if you could take a look, that'd be appreciated.

wisechengyi:
Filed a separate issue for lifecycle issue: https://github.com/pantsbuild/intellij-pants-plugin/issues/444

PR #440, opened 2019-11-22
[Refactoring] Make the boolean fields in `PantsProjectSettings` public

blorente:
Depends on https://github.com/pantsbuild/intellij-pants-plugin/pull/438

The boolean fields in `PantsProjectSettings` all have identity accessors (`public boolean is<whatever>() { return myWhatever; }`) and mutators (setters), making them effectively public mutable variables.

If we make them public we can get rid of all the getters and setters, reducing some duplication in that file.

The last commit is the important one.


wisechengyi:
Code fail to build actually, reverted.

PR #445, opened 2019-11-26
Minor fixes on export dep as jar

wisechengyi:
Corresponding change in Pants: https://github.com/pantsbuild/pants/pull/8632


wisechengyi:
ah no worries :)

PR #446, opened 2019-11-26
Add experimental note on import deps as jars

wisechengyi:


blorente:
The CI red is the same issue we've been seeing in master with the leaked projects, so this is probably ready to merge.

PR #447, opened 2019-11-27
Add option to use IntelliJ compiler for Pants projects

wisechengyi:
### Problem
Often times shelling off to Pants has non trivial latency. This PR introduces using IntelliJ compiler for an imported Pants project to achieve faster turnaround time.

### Visible changes includes:
1. Add a checkbox at import time for IntelliJ Compiler
Screen Shot 2019-12-12 at 10 28 38 AM

Once selected, all the test configurations generated will invoke the default `Build` action as below
Screen Shot 2019-12-05 at 9 29 06 AM

instead of `PantsCompile`
Screen Shot 2019-12-05 at 9 28 58 AM

2. No longer hide the default `Build Project` button.
Screen Shot 2019-12-05 at 9 40 08 AM

3. Module content roots are further trimmed to the top level roots, so intellij won't double count the sources as it gathers all the sources for each content root. 
#### Before
In this case, `WelSpecXYZ` will be passed to the compiler twice, causing errors.
Screen Shot 2019-12-05 at 9 50 30 AM

#### After
Screen Shot 2019-12-05 at 9 51 26 AM


wisechengyi:
TODO: add and fix some tests.

src/com/twitter/intellij/pants/execution/PantsMakeBeforeRun.java
```diff
@@ -89,9 +90,12 @@ public static void replaceDefaultMakeWithPantsMake(@NotNull Project project, @No
     if (!(runManager instanceof RunManagerImpl)) {
       return;
     }
-    RunManagerImpl runManagerImpl = (RunManagerImpl) runManager;
     RunConfiguration runConfiguration = settings.getConfiguration();
+    replaceDefaultMakeWithPantsMake(project, runConfiguration);
+  }
+
 
+  public static void replaceDefaultMakeWithPantsMake(@NotNull Project project, @NotNull RunConfiguration runConfiguration) {
```
blorente: If I understand correctly, this operation noops if we have the IntelliJ compiler.
From the name of the function, I wouldn't expect it to noop in any case, so I'd either remove the conditional [here](https://github.com/pantsbuild/intellij-pants-plugin/pull/447/files#diff-77d87e704dfb9b6b30ffaa3a11f7d6cdR132) (and push it to the callsites), or (less prefereably) change the name to reflect that it may not do anything, leaving a comment doc.

src/com/twitter/intellij/pants/service/PantsCompileOptionsExecutor.java
```diff
@@ -215,6 +215,9 @@ private GeneralCommandLine getPantsExportCommand(final File outputFile, @NotNull
     commandLine.addParameter("--no-quiet");
     if (getOptions().isImportSourceDepsAsJars()) {
       commandLine.addParameter("export-dep-as-jar");
+      if (myResolveSourcesAndDocsForJars) {
```
blorente: I'm not sure users mean this when they use this settings.
Namely, I'd think right now users expect this setting to download sources for their 3rdparty dependencies, and they'd expect sources to be accessible for their source dependencies by default (as is the status quo).

I'd be more comfortable with adding another toggle saying "I want sources for my source deps", which can even be on the same line, like:
` [ ] Import source deps as jars [ ] make sources accessible`
I feel like we'd want to turn this toggle on by default at the beginning to keep the UX as close to the original as possible, and tell people to maybe disable it in the future (whereas we may not want to toggle `resolveSourcesAnsDocsForJars` on by default ever for 3rdparty deps).

src/com/twitter/intellij/pants/service/project/PantsSystemProjectResolver.java
```diff
@@ -83,7 +83,9 @@
     task2executor.put(id, executor);
     final DataNode<ProjectData> projectDataNode =
       resolveProjectInfoImpl(id, executor, listener, isPreviewMode, settings.isEnableIncrementalImport());
-    doViewSwitch(id, projectPath);
+    if (!settings.isImportSourceDepsAsJars()) {
```
blorente: (From anecdotal evidence, take with a grain of salt)
The users I've talked to said that even with the improved "Project" view that "Import Source Deps as Jars" brings, they'd still prefer the "Project File Tree" view to be the default, as it allows them to search for files that are excluded more easily.

src/com/twitter/intellij/pants/service/project/model/TargetInfo.java
```diff
@@ -157,7 +157,9 @@ public boolean isJarLibrary() {
   }
 
   public boolean isScalaTarget() {
-    return getAddressInfos().stream().anyMatch(TargetAddressInfo::isScala);
+    return getAddressInfos().stream().anyMatch(TargetAddressInfo::isScala) ||
+           // TODO(yic): have Pants export `pants_target_type` correctly
+           getAddressInfos().stream().anyMatch(s -> s.getTargetAddress().endsWith("-scala"));
```
blorente: This is for Thrift targets, right? If so, I'd add an `NB:` explaining it.

src/com/twitter/intellij/pants/execution/PantsMakeBeforeRun.java
wisechengyi: Yeah good point. It does a bit more than change the build process to `PantsCompile` and changes the working directory to build root. So it will separate them out and wrap the logic of changing build process to a conditional.

src/com/twitter/intellij/pants/service/project/PantsSystemProjectResolver.java
wisechengyi: Gotcha. will revert.

src/com/twitter/intellij/pants/service/PantsCompileOptionsExecutor.java
blorente: Sounds good :)

src/com/twitter/intellij/pants/execution/PantsMakeBeforeRun.java
blorente: Thanks!

src/com/twitter/intellij/pants/settings/PantsProjectSettingsControl.java
```diff
@@ -77,6 +79,12 @@ protected void fillExtraControls(@NotNull PaintAwarePanel content, int indentLev
     myUseIdeaProjectJdkCheckBox.setSelected(mySettings.useIdeaProjectJdk);
     myImportSourceDepsAsJarsCheckBox.setSelected(mySettings.importSourceDepsAsJars);
     myUseIntellijCompilerCheckBox.setSelected(mySettings.useIntellijCompiler);
+    LinkLabel<?> intellijCompilerHelpMessage = LinkLabel.create(PantsBundle.message("pants.settings.text.use.intellij.compiler.help.messasge"), new Runnable() {
```
blorente: I can't quite see this message in the UI, it doesn't show up. Where should I be looking?

src/com/twitter/intellij/pants/settings/PantsSettings.java
```diff
@@ -76,6 +76,10 @@ public boolean isEnableIncrementalImport() {
     return getLinkedProjectsSettings().stream().anyMatch(PantsProjectSettings::isEnableIncrementalImport);
   }
 
+  public boolean isUseIntellijCompiler() {
+    return getLinkedProjectsSettings().stream().anyMatch(s -> s.useIntellijCompiler);
+  }
```
blorente: Does this mean that if any of my imported projects use the intellij compiler, all of them have to? I'm not sure what getLinkedProjectSettings does.

src/com/twitter/intellij/pants/settings/PantsProjectSettingsControl.java
wisechengyi: Updated the description under 1).

src/com/twitter/intellij/pants/settings/PantsSettings.java
wisechengyi: One IntelliJ window can contain multiple projects actually. So yes, that's what this means. 

This mostly just follows the pattern there. and the tricky part is that if a source file can belong to multiple projects, then which compiler should we use? I think this should be an okay heuristics.

PR #449, opened 2019-12-02
Project lifecycle issues

### LINKED ISSUE #444: project lifecycle issue
https://gist.github.com/wisechengyi/eab3389ed28934bd9e7c82e0a34ca82d

tpasternak:
Fix #444: two project lifecycle issues

1. Null Pointer Exception when closing IntelliJ if no projects has been opened
2. "Already disposed" exception in tests' teardown


tpasternak:
I'm still not sure if replace of `invokeLater` to `invokeAndWait` is the best solution, but it's the only one that is safe and does not require adding locking logic in many places.

Issue #355, opened 2018-06-11
BUILD file symbol resolve


wisechengyi:
currently pants dictionary keywords are highlighted by python plugin as unresolved symbols, so they need to be recognize correctly first before further assistance can be provided.

wisechengyi:
This would involve telling calling the below command to get the list of target types (may need a little change on the pants to make it more machine readable), and then feed the definition to python plugin.
```
$ ./pants targets
Scrubbed PYTHONPATH=/Users/yic/workspace/pants/src/python: from the environment.
                         alias: A target that gets replaced by its dependencies.
          annotation_processor: A Java library containing annotation processors.
                     benchmark: A caliper benchmark.
                contrib_plugin: A contributed pants plugin published by pantsbuild.
                    cpp_binary: A C++ binary.
                   cpp_library: A statically linked C++ library.
                   credentials: <Add description>
   ctypes_compatible_c_library: <Add description>
 ctypes_compatible_cpp_library: <Add description>
       external_native_library: A set of Conan package strings to be passed to the Conan package manager.
                         files: A collection of loose files.
                     go_binary: A local Go main package.
                    go_library: A local Go package.
           go_protobuf_library: A Go library generated from Protobuf IDL files.
           go_remote_libraries: A remote Go package.
             go_remote_library: A remote Go package.
             go_thrift_library: A Go library generated from Thrift IDL files.
                   jar_library: A set of external JAR files.
                    java_agent: A Java agent entrypoint.
            java_antlr_library: A Java library generated from Antlr grammar files.
             java_avro_library: Defines a target that builds Java code from Avro schema, protocol, or IDL files.
                  java_library: A Java library.
         java_protobuf_library: A Java library generated from Protocol Buffer IDL files.
            java_ragel_library: A Java library generated from a Ragel file.
           java_thrift_library: A Java library generated from Thrift IDL files.
          java_thrifty_library: An Android-optimized Java library generated by the Microsoft Thrifty thrift compiler.
             java_wire_library: A Java library generated from Wire IDL files.
                  javac_plugin: A Java compiler plugin.
                jax_ws_library: Generates a Java library from JAX-WS wsdl files.
                  jaxb_library: A Java library generated from JAXB xsd files.
                   junit_tests: JUnit tests.
                       jvm_app: A deployable JVM application.
                    jvm_binary: A JVM binary.
              jvm_prep_command: A command (defined in a Java target) that must be run before other tasks in a goal.
[...]

PR #454, opened 2019-12-03
Resolve names inside BUILD files to take into account available target types

tgodzik:
Previously, all BUILD files would have functions like `scala_library` or `java_library` underlined red despite them being valid.

Now, we additionally query while exporting for available target types and save them to the `PropertiesComponent`.

This will work with newest pants version, so we might need to wait until there is nightly release. Not sure what is the procedure for changes both to pants and this plugin.

Fixes https://github.com/pantsbuild/intellij-pants-plugin/issues/355

### DIFF
--- a/common/com/twitter/intellij/pants/util/PantsConstants.java
+++ b/common/com/twitter/intellij/pants/util/PantsConstants.java
@@ -9,9 +9,6 @@ import com.intellij.openapi.externalSystem.model.ProjectSystemId;
 import com.twitter.intellij.pants.model.JdkRef;
 import org.jetbrains.annotations.NotNull;
 
-import java.util.List;
-
-
 public class PantsConstants {
   public static final String PANTS = "pants";
   public static final String PANTS_CONSOLE_NAME = "Pants Console";
--- a/common/com/twitter/intellij/pants/util/PantsConstants.java
+++ b/common/com/twitter/intellij/pants/util/PantsConstants.java
@@ -9,6 +9,8 @@ import com.intellij.openapi.externalSystem.model.ProjectSystemId;
 import com.twitter.intellij.pants.model.JdkRef;
 import org.jetbrains.annotations.NotNull;
 
+import java.util.List;
+
 
 public class PantsConstants {
   public static final String PANTS = "pants";
@@ -43,6 +45,7 @@ public class PantsConstants {
   // Used to initialize project sdk therefore use project processing weight, i.e, the highest.
   public static final Key<JdkRef> SDK_KEY = Key.create(JdkRef.class, ProjectKeys.PROJECT.getProcessingWeight());
 
+  public static final String PANTS_AVAILABLE_TARGETS_KEY = "available_targets";
   public static final String PANTS_CLI_OPTION_EXPORT_OUTPUT_FILE = "--export-output-file";
   public static final String PANTS_CLI_OPTION_LIST_OUTPUT_FILE = "--list-output-file";
   public static final String PANTS_CLI_OPTION_EXPORT_CLASSPATH_MANIFEST_JAR = "--export-classpath-manifest-jar-only";
--- a/src/com/twitter/intellij/pants/service/PantsCompileOptionsExecutor.java
+++ b/src/com/twitter/intellij/pants/service/PantsCompileOptionsExecutor.java
@@ -196,8 +196,22 @@ public class PantsCompileOptionsExecutor {
     return processOutput;
   }
 
+  private boolean isAtLeastPantsVersion(int major, int minor) throws ExecutionException {
+    String pantsVersion = getPantsVersion();
+    String[] splitVersion = pantsVersion.trim().split("\\.");
+    return splitVersion.length >= 3 && Integer.parseInt(splitVersion[0]) >= major && Integer.parseInt(splitVersion[1]) >= minor;
+  }
+
   @NotNull
-  private GeneralCommandLine getPantsExportCommand(final File outputFile, @NotNull Consumer<String> statusConsumer) throws IOException {
+  private String getPantsVersion() throws ExecutionException {
+    final GeneralCommandLine commandLine = PantsUtil.defaultCommandLine(getProjectPath());
+    commandLine.addParameter("--version");
+    return getProcessOutput(commandLine).getStdout();
+  }
+
+  @NotNull
+  private GeneralCommandLine getPantsExportCommand(final File outputFile, @NotNull Consumer<String> statusConsumer)
+    throws IOException, ExecutionException {
     final GeneralCommandLine commandLine = PantsUtil.defaultCommandLine(getProjectPath());
 
     // Grab the import stage pants rc file for IntelliJ.
@@ -213,6 +227,11 @@ public class PantsCompileOptionsExecutor {
     }
     commandLine.addParameter("--target-spec-file=" + targetSpecsFile.getPath());
     commandLine.addParameter("--no-quiet");
+
+    boolean exportAvailableTypes = isAtLeastPantsVersion(1, 23);
+    if (exportAvailableTypes) {
+      commandLine.addParameter("--export-available-target-types");
+    }
     if (getOptions().isImportSourceDepsAsJars()) {
       commandLine.addParameter("export-dep-as-jar");
     }
--- a/src/com/twitter/intellij/pants/service/PantsCompileOptionsExecutor.java
+++ b/src/com/twitter/intellij/pants/service/PantsCompileOptionsExecutor.java
@@ -20,6 +20,7 @@ import com.twitter.intellij.pants.metrics.PantsMetrics;
 import com.twitter.intellij.pants.model.IJRC;
 import com.twitter.intellij.pants.model.PantsCompileOptions;
 import com.twitter.intellij.pants.model.PantsExecutionOptions;
+import com.twitter.intellij.pants.model.PantsOptions;
 import com.twitter.intellij.pants.settings.PantsExecutionSettings;
 import com.twitter.intellij.pants.util.PantsUtil;
 import org.jetbrains.annotations.Nls;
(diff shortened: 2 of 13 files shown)


tgodzik:
@wisechengyi anything else I can take a look at while working on this? Seems `blop` is still highlighted as error, can we get all possible functions from somewhere?

tgodzik:
Also I imagine we can add the target types to have completions here: PantsCompletionContributor 
Does that sound ok?

src/com/twitter/intellij/pants/service/PantsCompileOptionsExecutor.java
```diff
@@ -196,8 +196,22 @@ private ProcessOutput getProcessOutput(
     return processOutput;
   }
 
+  private boolean isAtLeastPantsVersion(int major, int minor) throws ExecutionException {
+    String pantsVersion = getPantsVersion();
+    String[] splitVersion = pantsVersion.trim().split("\\.");
+    return splitVersion.length >= 3 && Integer.parseInt(splitVersion[0]) >= major && Integer.parseInt(splitVersion[1]) >= minor;
+  }
```
wisechengyi: Could you make this a wrapper function of https://github.com/pantsbuild/intellij-pants-plugin/blob/e74ce17b082979b9d0982956ed5ef58b9f5a4985/common/com/twitter/intellij/pants/util/PantsUtil.java#L916, or possibly directly use the other instead?

src/com/twitter/intellij/pants/service/PantsCompileOptionsExecutor.java
```diff
@@ -213,6 +227,11 @@ private GeneralCommandLine getPantsExportCommand(final File outputFile, @NotNull
     }
     commandLine.addParameter("--target-spec-file=" + targetSpecsFile.getPath());
     commandLine.addParameter("--no-quiet");
+
+    boolean exportAvailableTypes = isAtLeastPantsVersion(1, 23);
```
wisechengyi: let's make it >= 1.24

src/com/twitter/intellij/pants/service/PantsCompileOptionsExecutor.java
```diff
@@ -196,8 +196,22 @@ private ProcessOutput getProcessOutput(
...
+    String[] splitVersion = pantsVersion.trim().split("\\.");
+    return splitVersion.length >= 3 && Integer.parseInt(splitVersion[0]) >= major && Integer.parseInt(splitVersion[1]) >= minor;
+  }
+
+  @NotNull
+  private String getPantsVersion() throws ExecutionException {
+    final GeneralCommandLine commandLine = PantsUtil.defaultCommandLine(getProjectPath());
+    commandLine.addParameter("--version");
+    return getProcessOutput(commandLine).getStdout();
+  }
```
wisechengyi: You can grab this from `PantsOptions` with `pants_version` as the key
```
$ ./pants options | grep pants_version
pants_version = 1.24.0.dev1 (from HARDCODED)
```
So PantsOptions should be cached already at this point, so that will save a call to Pants.

src/com/twitter/intellij/pants/service/PantsCompileOptionsExecutor.java
```diff
@@ -213,6 +227,11 @@ private GeneralCommandLine getPantsExportCommand(final File outputFile, @NotNull
     }
     commandLine.addParameter("--target-spec-file=" + targetSpecsFile.getPath());
     commandLine.addParameter("--no-quiet");
+
+    boolean exportAvailableTypes = isAtLeastPantsVersion(1, 23);
+    if (exportAvailableTypes) {
```
wisechengyi: can inline the variable probably.

tests/com/twitter/intellij/pants/integration/TargetFileResolutionIntegrationTest.java
```diff
@@ -0,0 +1,32 @@
...
+import com.intellij.psi.PsiManager;
+import com.intellij.psi.PsiReference;
+import com.twitter.intellij.pants.testFramework.OSSPantsIntegrationTest;
+
+import java.io.IOException;
+import java.util.Collection;
+
+public class TargetFileResolutionIntegrationTest extends OSSPantsIntegrationTest {
+
+  public void testAvailableTargetTypes() throws IOException {
```
tgodzik: Need to check whether we can run the test depending on the pants version still.

src/com/twitter/intellij/pants/service/project/model/ProjectInfo.java
```diff
@@ -40,6 +43,9 @@ public ProjectInfo() {
   // name to info
   protected Map<String, TargetInfo> targets;
 
+  @SerializedName("available_target_types")
+  protected String[] availableTargetTypes = {};
```
wisechengyi: not an issue right now, but if we want to expand the features a little to include what fields a target type should include, e.g.
```
java_library( 
 dependencies = [...]
)
```

It might need to suggest that `java_libraray` should have `dependencies` as a sub field and list of string as its type.

Something to keep in mind, or maybe add as a comment.

tests/com/twitter/intellij/pants/integration/TargetFileResolutionIntegrationTest.java
```diff
@@ -0,0 +1,32 @@
...
+    doImport("examples/src/scala/org/pantsbuild/example/hello/");
+    VirtualFile vfile  = myProjectRoot.findFileByRelativePath("BUILD");
+    assertNotNull(vfile);
+    PsiFile build = PsiManager.getInstance(myProject).findFile(vfile);
+    String input = new String(myProjectRoot.findFileByRelativePath("BUILD").contentsToByteArray());
```
wisechengyi: `myProjectRoot.findFileByRelativePath("BUILD")` -> `vfile`?

tests/com/twitter/intellij/pants/integration/OSSProjectInfoResolveTest.java
```diff
@@ -63,6 +65,15 @@ public void testTargetType() {
     assertFalse(greetTarget.isScalaTarget());
   }
 
+  public void testAvailableTargetTypes() {
+    final ProjectInfo info = resolveProjectInfo("examples/src/scala/org/pantsbuild/example/hello/");
+
+    final List<String> availableTargetTypes = Arrays.asList(info.getAvailableTargetTypes());
+    assertNotNull(availableTargetTypes);
```
wisechengyi: not null should be asserted on `info.getAvailableTargetTypes()` instead of `availableTargetTypes`, or you can remove this assert probably, because `Arrays.asList` will probably fail on null anyway.

tests/com/twitter/intellij/pants/integration/TargetFileResolutionIntegrationTest.java
```diff
@@ -0,0 +1,32 @@
...
+    final PsiReference reference = build.findReferenceAt(input.indexOf("files(") + 1);
```
wisechengyi: Is this testing this file?
```
$ cat examples/src/scala/org/pantsbuild/example/hello/BUILD 
#This is a dummy build file for Intellij Plugin integration test.
#Please do not remove this BUILD file

target(
  dependencies = [
   'examples/src/scala/org/pantsbuild/example/hello/exe:exe',
  ]
)
```

if so where's `files(` coming from?

common/com/twitter/intellij/pants/util/PantsConstants.java
```diff
@@ -9,6 +9,8 @@
 import com.twitter.intellij.pants.model.JdkRef;
 import org.jetbrains.annotations.NotNull;
 
+import java.util.List;
```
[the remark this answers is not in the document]
tgodzik: Removed

src/com/twitter/intellij/pants/service/PantsCompileOptionsExecutor.java
tgodzik: Fixed! I added a new method `isCompatibleVersion` - it's a bit easier to reason about than when returning int I think :thinking:

src/com/twitter/intellij/pants/service/PantsCompileOptionsExecutor.java
tgodzik: Done!

src/com/twitter/intellij/pants/service/PantsCompileOptionsExecutor.java
tgodzik: Fixed

src/com/twitter/intellij/pants/service/PantsCompileOptionsExecutor.java
tgodzik: Sure, I inlined it

src/com/twitter/intellij/pants/service/project/model/ProjectInfo.java
tgodzik: Added a comment about it - should be resonable to do in the future here.

tests/com/twitter/intellij/pants/integration/TargetFileResolutionIntegrationTest.java
tgodzik: Forgot about that one, thanks!

tests/com/twitter/intellij/pants/integration/OSSProjectInfoResolveTest.java
tgodzik: Removed that and added `@NotNull` annotation - I think Gson just defaults to an empty array but just to be on thesafe side.

tests/com/twitter/intellij/pants/integration/TargetFileResolutionIntegrationTest.java
tgodzik: it was actually the root `BUILD` file - changed that to use the same file as the one in import.

tgodzik:
@wisechengyi Thanks for the review! It clarified quite a bit of things. I think I fixed all the issues, could you take a second look at it?

testFramework/com/twitter/intellij/pants/testFramework/OSSPantsIntegrationTest.java
```diff
@@ -63,7 +63,7 @@ protected void assertEmptyBeforeRunTask(RunConfiguration configuration) {
    * Assert Project has the right JDK and language level (JVM project only).
    */
   protected void assertProjectJdkAndLanguageLevel() {
-    final String pantsExecutablePath = PantsUtil.findPantsExecutable(getParentPath()).get().getPath();
+    final String pantsExecutablePath = PantsUtil.findPantsExecutable(getProjectPath()).get().getPath();
```
tgodzik: This used the pants inside of `pants-intellij-plugin` instead of the one from the project. Export was failing due to:
```
FAILURE: No bootstrap callback registered for scalac in scala
```
It still fails if I run `./pants export` from the intellij plugin pants, any idea what it's about?

tests/com/twitter/intellij/pants/model/SimpleExportResultTest.java
```diff
@@ -100,12 +100,12 @@ public void testParseExport_1_0_6() throws Exception {
   }
 
   public void testExportCache() {
-    SimpleExportResult export_a = SimpleExportResult.getExportResult("./pants");
-    SimpleExportResult export_b = SimpleExportResult.getExportResult("./pants");
+    SimpleExportResult export_a = SimpleExportResult.getExportResult("./.cache/pants/pants");
```
tgodzik: Same thing as above, the command was failing due to:
```
FAILURE: No bootstrap callback registered for scalac in scala
```

testFramework/com/twitter/intellij/pants/testFramework/OSSPantsIntegrationTest.java
wisechengyi: Unclear. But thanks for fixing this!

PR #456, opened 2019-12-10
Make sure we don't offer project refresh on files in .pants.d directory

tgodzik:
Inside ` .pants.d ` there are build.html files, which are modified after each refresh causing infinite loop for refreshes.

### DIFF
--- a/src/com/twitter/intellij/pants/file/FileChangeTracker.java
+++ b/src/com/twitter/intellij/pants/file/FileChangeTracker.java
@@ -19,7 +19,9 @@ import com.intellij.openapi.vfs.VirtualFileManager;
 import com.intellij.openapi.vfs.VirtualFileMoveEvent;
 import com.intellij.openapi.vfs.VirtualFilePropertyEvent;
 import com.twitter.intellij.pants.metrics.PantsExternalMetricsListenerManager;
+import com.twitter.intellij.pants.model.PantsOptions;
 import com.twitter.intellij.pants.settings.PantsSettings;
+import com.twitter.intellij.pants.util.PantsConstants;
 import com.twitter.intellij.pants.util.PantsUtil;
 import org.apache.commons.lang3.StringUtils;
 import org.jetbrains.annotations.NotNull;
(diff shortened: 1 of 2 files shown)


tgodzik:
@wisechengyi This is a fix for the refresh issue in pants projects. Let me know if it's sensible!

src/com/twitter/intellij/pants/file/FileChangeTracker.java
```diff
@@ -129,7 +129,8 @@ private static ChangeType detectChangeType(Project project, @NotNull VirtualFile
     if (rootManager.getFileIndex().getContentRootForFile(file) != null) {
       boolean isBuildFile = PantsUtil.isBUILDFileName(file.getName());
       boolean isUnderPantsRepo = PantsUtil.isFileUnderPantsRepo(file);
-      if (isUnderPantsRepo && isBuildFile) {
+      boolean isNotInPantsD = !file.getPath().contains("/.pants.d/");
```
wisechengyi: Can we somehow use the `pants_workdir` from PantsOptions, instead of harcoding `/.pants.d/` because the workdir (default '.pants.d/') can actually be configured?

nit: can we use `isInPantsD` then use `!isInPantsD` in the if statement.

tgodzik: Fixed - I use `pants_workdir` as the default and checking for `/.pants.d/` as a fallback. It seems though that `pants_workdir` is always defined, so the fallback might not be actually needed.

src/com/twitter/intellij/pants/file/FileChangeTracker.java
```diff
@@ -129,7 +131,15 @@ private static ChangeType detectChangeType(Project project, @NotNull VirtualFile
     if (rootManager.getFileIndex().getContentRootForFile(file) != null) {
       boolean isBuildFile = PantsUtil.isBUILDFileName(file.getName());
       boolean isUnderPantsRepo = PantsUtil.isFileUnderPantsRepo(file);
-      if (isUnderPantsRepo && isBuildFile) {
+
+      Path filePath = Paths.get(file.getPath());
+      boolean shouldBeIgnored =
+        PantsOptions.getPantsOptions(project)
+          .flatMap(options -> options.get("pants_workdir"))
```
wisechengyi: can we move `pants_workdir` to a constant? https://github.com/pantsbuild/intellij-pants-plugin/blob/master/common/com/twitter/intellij/pants/util/PantsConstants.java#L54

tgodzik: Fixed!

PR #458, opened 2019-12-11
Don't drop list of targets while saving settings

### LINKED ISSUE #457: Cannot reimport project after settings change
Intellij 2019.3 
Pants plugin commit c44af982f22b84bd45300788b3f586ca3f591f50
Repro steps:
1. Import project with "Enable incremental import" option checked
2. Wait for the import process to finish
3. Open "File" -> "Settings" -> "Build, Execution, Deployment" -> "Pants"
4. Uncheck "Enable incremental Project Import ..." checkbox
5. Click "Apply"
6. Check "Enable incremental Project Import ..." checkbox again
7. Clock "OK"
8. "Pants" -> "Refresh Pants Project"

Expected result:
Succsessful reimport of the project

Actual result:
Message: "No target roots found in build graph. Please make sure Pants export version >= 1.0.9"

reimport_failure

tpasternak:
Before this commit, the list of the targets were stored in
`PantsProjectSettings::myTargetSpecs` field. Unfortunately, this list was based on the
contents of `myTargetSpecsBox` widget, which was cleared each time
when the "Pants->Settings" window was opened. As a result, the list
of targets was cleared each time during settings save.

The fix to this is just not to clean the list of targets in the
window.

But this led to a new issue. When the list of targets is presented
to the user, it's also possible for him to deselect target. During
save, the `myTargetSpecs` variable was based on the list of selected
targets, so after deselection and save, it was not possible to select
it again.

In order to overcome this, I put two lists to `PantsProjectSettings`
instead of `myTargetSpecs`:

 1. `allAvailableTargetSpecs` - used to build `myTargetSpecsBox` widget
 2. `selectedTargetSpecs` - to inform `pants` which targets it should
operate on

fixes #457

### DIFF
--- a/common/com/twitter/intellij/pants/model/PantsExecutionOptions.java
+++ b/common/com/twitter/intellij/pants/model/PantsExecutionOptions.java
@@ -10,7 +10,7 @@ import java.util.List;
 public interface PantsExecutionOptions {
 
   @NotNull
-  List<String> getTargetSpecs();
+  List<String> getSelectedTargetSpecs();
 
   boolean isEnableIncrementalImport();
   boolean isImportSourceDepsAsJars();
--- a/common/com/twitter/intellij/pants/util/PantsConstants.java
+++ b/common/com/twitter/intellij/pants/util/PantsConstants.java
@@ -34,6 +34,7 @@ public class PantsConstants {
   public static final String PANTS_LIBRARY_EXCLUDES_KEY = "pants.library.excludes";
   public static final String PANTS_TARGET_ADDRESSES_KEY = "pants.target.addresses";
   public static final String PANTS_TARGET_ADDRESS_INFOS_KEY = "pants.target.address.infos";
+  public static final String PANTS_TARGET_MODULE_TYPE =  "pants.module";
 
   public static final String PANTS_OPTION_PANTS_WORKDIR = "pants_workdir";
   public static final String PANTS_OPTION_TEST_JUNIT_STRICT_JVM_VERSION = "test.junit.strict_jvm_version";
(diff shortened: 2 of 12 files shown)


tpasternak:
Thanks for the comment. 

> When we change the targets in the settings, pants does not auto-refresh or notify user to refresh the targets. Would be good at least notify. We can do this as a separate PR.

I would prefer to put it into a separate PR, because there was no auto-refresh before as well.

>     2. we allow user to deselect all targets?
>         If I deselect all, click `Ok` and refresh pants projects, IntelliJ shows a notification with

This happens because the orphan modules detection is done as a `onSuccessImport` action in `AbstractModuleDataService`. I haven't found way to disable it yet. But I'm not sure whether we really want to disable it. The other soultion would be to disable modules deselection in Pants Options. De facto it was disabled before, because the checkbox widget was hidden. Would it be OK for you?

tpasternak:
Ok, IntelliJ doesn't try to reload orphan modules any more.

Sohamr:
> Thanks for the comment.
> 
> > When we change the targets in the settings, pants does not auto-refresh or notify user to refresh the targets. Would be good at least notify. We can do this as a separate PR.
> 
> I would prefer to put it into a separate PR, because there was no auto-refresh before as well.
Yes, agree with having a separate PR.
> > ```
> > 2. we allow user to deselect all targets?
> >     If I deselect all, click `Ok` and refresh pants projects, IntelliJ shows a notification with
> > ```
> 
> This happens because the orphan modules detection is done as a `onSuccessImport` action in `AbstractModuleDataService`. I haven't found way to disable it yet. But I'm not sure whether we really want to disable it. The other soultion would be to disable modules deselection in Pants Options. De facto it was disabled before, because the checkbox widget was hidden. Would it be OK for you?

Thanks for fixing this. We can do any remaining changes in separate PR.

PR #460, opened 2019-12-11
Bump pants in dummy repo

wisechengyi:


testData/dummy_repo/pants
```diff
@@ -3,100 +3,191 @@
...
 # ====================================================================
 
-set -e
+set -eou pipefail
 
-PYTHON=${PYTHON:-$(which python2.7)}
+PYTHON_BIN_NAME="${PYTHON:-unspecified}"
 
-PANTS_HOME="${PANTS_HOME:-${HOME}/.cache/pants/setup}"
+PANTS_HOME="${PANTS_HOME:-${XDG_CACHE_HOME:-$HOME/.cache}/pants/setup}"
```
blorente: I'm very curious about how we ran into the need to specify `XDG_CACHE_HOME` here :) (not saying it's bad, it's a good change, I'm just very curious

testData/dummy_repo/pants
```diff
@@ -3,100 +3,191 @@
...
+
+function get_pants_ini_config_value {
+  if [[ ! -f 'pants.ini' ]]; then
+    return 0
+  fi
+  config_key="$1"
+  valid_delimiters="[:=]"
+  optional_space="[[:space:]]*"
+  prefix="^${config_key}${optional_space}${valid_delimiters}${optional_space}"
+  sed -ne "/${prefix}/ s#${prefix}##p" pants.ini
```
blorente: This is assuming the option name is given fully, right? as in, if `config_key` is something like `"cache.test.pytest"`, then the output will be multiline.
Should we error in those cases, or add a comment explaining that we want the fully qualified option name?

testData/dummy_repo/pants
```diff
@@ -3,100 +3,191 @@
...
+    curl -sSL https://pypi.python.org/pypi/pantsbuild.pants/json |
+      python -c "import json, sys; print(json.load(sys.stdin)['info']['version'])"
+  fi
+}
+
+function determine_default_python_exe {
+  pants_version="$1"
+  pants_major_version="$(echo "${pants_version}" | cut -d '.' -f1)"
+  pants_minor_version="$(echo "${pants_version}" | cut -d '.' -f2)"
+  if [[ "${pants_major_version}" -eq 1 && "${pants_minor_version}" -le 14 ]]; then
```
blorente: This list seems like it would grow pretty stale pretty fast.
It might be some work, but could this `Map[PantsMinorVersion, Set[PythonVersion]]` live somewhere in the pants repo that we can parse, or download? Otherwise we might be stuck manually updating this list whenever we do a major bump in python versions we support (like if we stop supporting 3.6).

That said, it might be more work than it's worth it, so OK to leave it as a TODO, or a note explaining that we don't plan on deprecating 3.6 at all.

testData/dummy_repo/pants
wisechengyi: @blorente Thanks for the review! Sorry I didn't provide enough context. The script was from the installation command from https://www.pantsbuild.org/install.html. Do you think we can land as is?

wisechengyi: The related code is under https://github.com/pantsbuild/setup/blob/gh-pages/pants

blorente: Oh cool, yeah :) that makes sense.

PR #466, opened 2019-12-17
Create additional inspection to allow to automatically add Python Facets to modules

tgodzik:
All BUILD files require a python interpreter and if a module doesn't contain one, there will be a warning shown unless the inspection is disabled. Now, we add an actionable quickfix that will automatically add a Python facet to a module in order to not replace a possible Java Jdk set already.

Overall quickfix algorithm:
1. If there is no python interpreter for the project, a window pops up where the user can choose the python interpreter to use. Otherwise skip.
2. If there is an already existing empty facet we remove it.
3. We add a new facet with either and existing interpreter or the one chosen in 1. step.

2. is needed since just adding an SDK is a flaky and I wasn't able to make it visible in the UI.

facets

### DIFF
--- a/resources/META-INF/plugin.xml
+++ b/resources/META-INF/plugin.xml
@@ -120,6 +120,9 @@
 
   <extensions defaultExtensionNs="com.intellij">
     <macro implementation="com.twitter.intellij.pants.macro.FilePathRelativeToBuiltRootMacro"/>
+    <localInspection bundle="com.twitter.intellij.pants.PantsBundle" key="pants.info.python.facet.missing"
+                     groupName="Pants" enabledByDefault="true"
+                     implementationClass="com.twitter.intellij.pants.inspection.PythonFacetInspection"/>
     <localInspection bundle="com.twitter.intellij.pants.PantsBundle" key="pants.info.python.plugin.missing"
                      groupName="Pants" enabledByDefault="true"
                      implementationClass="com.twitter.intellij.pants.inspection.PythonPluginInspection"/>
--- a/resources/META-INF/plugin.xml
+++ b/resources/META-INF/plugin.xml
@@ -120,9 +120,6 @@
 
   <extensions defaultExtensionNs="com.intellij">
     <macro implementation="com.twitter.intellij.pants.macro.FilePathRelativeToBuiltRootMacro"/>
-    <localInspection bundle="com.twitter.intellij.pants.PantsBundle" key="pants.info.python.facet.missing"
-                     groupName="Pants" enabledByDefault="true"
-                     implementationClass="com.twitter.intellij.pants.inspection.PythonFacetInspection"/>
     <localInspection bundle="com.twitter.intellij.pants.PantsBundle" key="pants.info.python.plugin.missing"
                      groupName="Pants" enabledByDefault="true"
                      implementationClass="com.twitter.intellij.pants.inspection.PythonPluginInspection"/>
--- a/common/com/twitter/intellij/pants/PantsBundle.properties
+++ b/common/com/twitter/intellij/pants/PantsBundle.properties
@@ -11,6 +11,7 @@ pants.library.no.pex.file=Couldn't find pex file for version ''{0}''
 pants.title.project.files=Project Files Tree
 
 pants.info.python.plugin.missing=Please install Python Support plugin for code assistance in BUILD files.
+pants.info.python.facet.missing=Automatically add Python interpreter for a module containing this BUILD file.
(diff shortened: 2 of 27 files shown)


tgodzik:
@wisechengyi @Sohamr  This should be ready to be reviewed.

wisechengyi:
Thanks, @tgodzik ! Is there some way we can guide the user to user system interpreter by default? Otherwise user might accidentally add a virtualenv there. E.g. https://drive.google.com/open?id=1OA5BT0Uh2qLWIRGfj_YYV2allGVcsV0y

wisechengyi:
or ideally we can automatically pick an interpreter.

tgodzik:
> or ideally we can automatically pick an interpreter.

I don't think we should choose automatically - especially that the user might not have the system interpreter.

> Thanks, @tgodzik ! Is there some way we can guide the user to user system interpreter by default? Otherwise user might accidentally add a virtualenv there. E.g. https://drive.google.com/open?id=1OA5BT0Uh2qLWIRGfj_YYV2allGVcsV0y

From what I see there is not actually way to select the default one. But will double check.

tgodzik:
> It asks to add a Python facet every time a BUILD file is opened. Once selected for a BUILD file can we auto-add it for all the BUILD files?

We should be able to add it to all modules, would that be alright?

> Getting below error, I think I remember seeing this before but not 100% sure.

I don't think this is related since it does mention Groovy 
```
	at org.jetbrains.plugins.groovy.compiler.GroovyTargetScopeProvider.lambda$containsGroovyResources$2(GroovyTargetScopeProvider.java:82)
```

marek1840:
The build seems to fail with an unrelated error:
```
22:34:23 00:01       [coursier]
                     ==== stdout ====
                     
                     ==== stderr ====
                     OpenJDK 64-Bit Server VM warning: ignoring option MaxPermSize=256m; support was removed in 8.0
                       Result:
                     └─ com.martiansoftware:nailgun-server:0.9.1
                     
                     Error:
                       com.martiansoftware:nailgun-server:0.9.1
                         No fallback URL found
                         download error: Caught java.io.IOException: Server returned HTTP response code: 403 for URL: https://repo1.maven.org/maven2/com/martiansoftware/nailgun-server/0.9.1/nailgun-server-0.9.1.pom (Server returned HTTP response code: 403 for URL: https://repo1.maven.org/maven2/com/martiansoftware/nailgun-server/0.9.1/nailgun-server-0.9.1.pom) while downloading https://repo1.maven.org/maven2/com/martiansoftware/nailgun-server/0.9.1/nailgun-server-0.9.1.pom
                     
FAILURE: The coursier process exited non-zero: 1
```

Sohamr:
> Tests seem to fail with an unrelated error:
> 
> ```
> 22:34:23 00:01       [coursier]
>                      ==== stdout ====
>                      
>                      ==== stderr ====
>                      OpenJDK 64-Bit Server VM warning: ignoring option MaxPermSize=256m; support was removed in 8.0
>                        Result:
>                      └─ com.martiansoftware:nailgun-server:0.9.1
>                      
>                      Error:
>                        com.martiansoftware:nailgun-server:0.9.1
>                          No fallback URL found
>                          download error: Caught java.io.IOException: Server returned HTTP response code: 403 for URL: https://repo1.maven.org/maven2/com/martiansoftware/nailgun-server/0.9.1/nailgun-server-0.9.1.pom (Server returned HTTP response code: 403 for URL: https://repo1.maven.org/maven2/com/martiansoftware/nailgun-server/0.9.1/nailgun-server-0.9.1.pom) while downloading https://repo1.maven.org/maven2/com/martiansoftware/nailgun-server/0.9.1/nailgun-server-0.9.1.pom
>                      
> FAILURE: The coursier process exited non-zero: 1
> ```

Agree, seems unrelated. Mind rebasing with latest master?

@wisechengyi might have more idea.

resources/META-INF/plugin.xml
```diff
@@ -120,6 +120,9 @@
 
   <extensions defaultExtensionNs="com.intellij">
     <macro implementation="com.twitter.intellij.pants.macro.FilePathRelativeToBuiltRootMacro"/>
+    <localInspection bundle="com.twitter.intellij.pants.PantsBundle" key="pants.info.python.facet.missing"
+                     groupName="Pants" enabledByDefault="true"
+                     implementationClass="com.twitter.intellij.pants.inspection.PythonFacetInspection"/>
```
wisechengyi: Hey might want to move this to `pants-python.xml` otherwise when python plugin is disabled, it will have the following error when clicking `Add python facet`:
Screen Shot 2020-02-10 at 11 10 31 AM

PR #455, opened 2019-12-09
Contribute available build target types for completions inside of BUILD files.

tgodzik:
This is a follow up for https://github.com/pantsbuild/intellij-pants-plugin/pull/454

Using the same information we are able to provide sensible completions for target types such as `scala_library` or `java_library`

pants

@wisechengyi Let me know if we need something more here.

It would be nice to also auto complete and highlight properties inside the functions, but I am not sure where to get that information properly. Most likely we would need to improve the structure inside of `pants export`

PR #473, opened 2020-01-08
Contribute available build target types for completions inside of BUILD files

lukaszwawrzyk:
This is a follow up for #455 as I took this task from @tgodzik 
I rebased this branch on top of master and addressed the comment from the original PR so that full addresses are used instead of paths. 

I replaced `PantsBuildFileIndex` which was only used for completions with `PantsAddressesIndex`, so now for all indexed BUILD files I take all of the target names from `PsiFile`s and add the default name also (equal to name of parent dir of BUILD file).

Initially I tried the suggested approach of reading all addresses from iml for all modules, but it made PantsCompletionTest#testAbsoluteAddress fail as it seems that project is not imported by this test properly.
Current approach gives wider set of target paths for completion (includes non-imported modules) and is more similar to the original behavior. Also it doesn't require project refresh to pickup new targets that were added.
If we'd rather prefer the smaller scope and relay on iml files, I can change the code, but I think I'd delete that single failing test as it would be quite an effort to fix it and it is also kind of covered by the new tests in this PR.

@wisechengyi Could you check if this PR is ok now?


src/com/twitter/intellij/pants/index/PantsAddressesIndex.java
```diff
@@ -0,0 +1,114 @@
...
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+public class PantsAddressesIndex extends ScalarIndexExtension<String> {
+  public static final ID<String, Void> NAME = ID.create("PantsAddressesIndex");
+
+  public static Collection<String> getAddresses(@NotNull PsiFile file) {
+    // We need to make sure that the files really belong to the project.
+    // As stated in http://www.jetbrains.org/intellij/sdk/docs/basics/indexing_and_psi_stubs/file_based_indexes.html#accessing-a-file-based-index
+    // it "may also contain additional keys not currently found in the project."
```
wisechengyi: Thanks for the doc!

PR #482, opened 2020-01-22
Workaround for "Find Usages broken" bug in Scala Plugin

tpasternak:
Bug tracker: https://youtrack.jetbrains.com/issue/SCL-16768

Correct initialization of `ScalaCompilerReferenceService` is required for "Find usages"                                                                   feature to work correcly. Unfortunately, it's broken in that way, it does nothing when                                                                    there are no modules in the project right after the moment, when the project is opened.                                                                   This happens when a project is imported, the module creation starts after opening then.

### DIFF
--- a/src/com/twitter/intellij/pants/service/scala/PantsScalaDataService.scala
+++ b/src/com/twitter/intellij/pants/service/scala/PantsScalaDataService.scala
@@ -52,16 +52,16 @@ class PantsScalaDataService extends ProjectDataService[ScalaModelData, Library] 
     // This happens when a project is imported, the module creation starts after opening then. 
     project.getMessageBus().connect().subscribe(ProjectTopics.MODULES, new ModuleListener {
       var done = false
+
       override def moduleAdded(project: Project,
-                               module: Module): Unit =  {
-        if(!done) {
+                               module: Module): Unit = {
+        if (!done) {
           if (PantsUtil.isPantsModule(module)) {
             ScalaCompilerReferenceService(project).projectOpened()
             done = true;
           }
         }
       }
-
     });
 
   }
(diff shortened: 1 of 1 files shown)


wisechengyi:
### 1


With this patch, it will get stuck at "indexing recompiled classfiles"

Repro:
* import `examples/src/scala/org/pantsbuild/example/hello`
* Apply your upstream repro:
```
case class MyClass(x: Int) {
  def apply(y: String) = 1
}

object MyClass {
  def instance = MyClass(1)("1")
}
```
in exe.scala

### 2
We should probably only enable this feature if IntelliJ compiler is used for the project, i.e. when `Use the IntelliJ Compiler` is checked during import.

 Example on fetching the setting: https://github.com/pantsbuild/intellij-pants-plugin/blob/5b983db1303b97d65a1c3fa1aa005197fd54c4e7/src/com/twitter/intellij/pants/components/impl/PantsProjectComponentImpl.java#L243-L246

### 3

If it won't work regardless, we can send a clear message to user this is not supported.

tpasternak:
Hi, thanks for the comments. I'm not going to merge it in this form, as we need to change scala plugin to guarantee that the `projectOpened` method is not called more than once. Currently, it's impossible. I'm waiting for the response from intellij-scala maintainers https://github.com/JetBrains/intellij-scala/pull/495

My answers:
> Repro:
>    import examples/src/scala/org/pantsbuild/example/hello
>    Apply your upstream repro:

Ok, this seems to be a separate issue, I don't know what's the root cause yet, but probably it's something related to `buildCompilationDiff` counter in `JpsCompilationWatcher`. It occurs on master as well (remember to close & open IntelliJ after import). Submitted here https://github.com/pantsbuild/intellij-pants-plugin/issues/484

> We should probably only enable this feature if IntelliJ compiler is used for the project, i.e. when Use the IntelliJ Compiler is checked during import.

That's true, I' take this into account when I will be prepraring the right soltution after doing fix/workaround in intelllij-scala

> If it won't work regardless, we can send a clear message to user this is not supported.

I don't know if it's possible, probably it would require changing intellij-scala as well.

PR #485, opened 2020-01-27
Workaround for "Find Usages broken" bug in Scala Plugin

lukaszwawrzyk:
Followup on https://github.com/pantsbuild/intellij-pants-plugin/pull/482 as I will work on completing that issue.  We will have to wait for this https://github.com/JetBrains/intellij-scala/pull/498 to be released first.


wisechengyi:
Hi, does it mean to replace #482? 

Also the compile error in CI seems legit.

tpasternak:
Yes, exactly. @lukaszwawrzyk took this task for me

PR #490, opened 2020-02-06
Fix pants version parsing

blorente:
### Problem

Internally at Twitter, we append some extra information to the pants version number. However, this means that the pants version parser doesn't work when trying to parse stable releases (`1.20.0` as opposed to something like `1.20.0rc1`)

### Solution
Create a proper model for a `PantsVersion`, and use a validated regex to parse it.

TODO:

- [X] Add tests for parsing: The test is already there (com.twitter.intellij.pants.util.PantsUtilTest#testIsCompatibleVersion)
- [X] Integrate the new class more. This is the last commit. Since we read the version from a file, there aren't any real fields to update.


common/com/twitter/intellij/pants/model/PantsVersion.java
```diff
@@ -0,0 +1,66 @@
...
+import java.util.regex.Pattern;
+
+public class PantsVersion {
+
+  private int major;
+  private int minor;
+  private int patch;
+  private PantsPatchVersionKind kind;
+
+  // Link to regular expression cehcking this: https://regex101.com/r/SUZCl6/2
```
cosmicexplorer:
```suggestion
  // Link to regular expression checking this: https://regex101.com/r/SUZCl6/2
```

tests/com/twitter/intellij/pants/util/PantsUtilTest.java
```diff
@@ -40,17 +31,19 @@ public void testIsPantsProjectFile() {
...
-    assertFalse(PantsUtil.isCompatibleVersion("2.33.44", "2.34.44"));
-    assertFalse(PantsUtil.isCompatibleVersion("1.34.44", "2.34.44"));
+    assertTrue(PantsUtil.isCompatiblePantsVersion("2.2.3", "1.2.3"));
+    assertTrue(PantsUtil.isCompatiblePantsVersion("1.4.4", "1.2.3"));
+    assertTrue(PantsUtil.isCompatiblePantsVersion("1.2.4", "1.2.3"));
+    assertTrue(PantsUtil.isCompatiblePantsVersion("1.2.4rc0", "1.2.3"));
+    assertTrue(PantsUtil.isCompatiblePantsVersion("1.2.4.dev0", "1.2.3"));
+    assertTrue(PantsUtil.isCompatiblePantsVersion("1.2.3", "1.2.3"));
+    assertTrue(PantsUtil.isCompatiblePantsVersion("1.2.3+git12345678", "1.2.3"));
+    assertTrue(PantsUtil.isCompatiblePantsVersion("1.2.3-git12345678", "1.2.3"));
```
cosmicexplorer: I might also add a test for a version like `1.24.0rc1-git2481dab2-md59a1c1708` as well, since we see that in our internal pants repo. I am pretty sure that the parsing mechanism you've developed here is only going to capture the relevant numbers, but in the case that we modify the parsing in the future, it might be good to have a complete example?

common/com/twitter/intellij/pants/util/PantsUtil.java
```diff
@@ -214,16 +211,17 @@ public static boolean isScalaRelatedTestRunConfiguration(RunConfiguration rc) {
...
       .flatMap(exec -> PantsOptions.getPantsOptions(exec.getPath()).get("pants_version"))
-      .map(version -> PantsUtil.isCompatibleVersion(version, minVersion))
+      .map(version -> PantsUtil.isCompatiblePantsVersion(version, minVersion))
       .orElse(false);
   }
 
-  public static boolean isCompatibleVersion(String current, String minimum) {
-    String currentVersion = current.replaceAll("rc.+", "").trim();
-    return versionCompare(currentVersion, minimum) >= 0;
+  public static boolean isCompatiblePantsVersion(String current, String minimum) {
```
cosmicexplorer: Good change!! Thank you for fixing up naming hygiene here!

common/com/twitter/intellij/pants/model/PantsVersion.java
cosmicexplorer: Btw, *fantastic* move to link to regex101.com -- thanks for thinking about this! This is exactly how we make regex a little more comprehensible.

common/com/twitter/intellij/pants/model/PantsVersion.java
```diff
@@ -0,0 +1,66 @@
...
+  private static final String PANTS_VERSION_FORMAT = "(\\d+)\\.(\\d+)\\.(\\d+)(rc\\d+|\\.dev\\d+)?";
```
cosmicexplorer:
```suggestion
  private static final String PANTS_VERSION_FORMAT = "^(\\d+)\\.(\\d+)\\.(\\d+)(rc\\d+|\\.dev\\d+)?";
```

common/com/twitter/intellij/pants/model/PantsVersion.java
```diff
@@ -0,0 +1,66 @@
...
+  final Pattern pants_version_pattern = Pattern.compile(PANTS_VERSION_FORMAT, Pattern.MULTILINE);
```
cosmicexplorer: Is there a reason we're using `Pattern.MULTILINE` here? I believe that would convert the above `^` into start-of-line instead of start-of-input, see: https://stackoverflow.com/a/6143347/2518889.

wisechengyi:
master is broken atm: https://github.com/pantsbuild/intellij-pants-plugin/issues/491

will merge once that's fixed. This is not urgent given we have a workaround internally.

tests/com/twitter/intellij/pants/util/PantsUtilTest.java
blorente: Added! Thanks

common/com/twitter/intellij/pants/model/PantsVersion.java
blorente: Changed, thanks! I like the `^` a lot, btw :) Nice catch.

PR #494, opened 2020-02-12
Bump CI against 2019.3.3

wisechengyi:


sundresh:
I assume this means in Pants 1.25.0, the `--target-spec-file` option change to `--spec-file` (or the latter was added as another way of specifying the same option). I'd like to accept this change, but I don't see how. Do you have to add me as a reviewer?

wisechengyi:
> I assume this means in Pants 1.25.0, the `--target-spec-file` option change to `--spec-file` (or the latter was added as another way of specifying the same option).

Ah this landed separately: https://github.com/pantsbuild/intellij-pants-plugin/pull/493

because `--target-spec-file` is deprecated.

PR #498, opened 2020-02-24
Idea 201.5259.13 support

marek1840:
The major obstacle was that scala plugin for 2020 is being compiled with jdk 11, hence we can no longer run the pants-plugin test suite with the jdk 8.
Our solution is to:
A) set the 'java11' platform for test targets(misc, jvm-ibtegration and so on)
B) force pants-plugin to use JDK 8 when running pants within test cases but use JDK 11 to run the test cases.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,10 +20,13 @@ addons:
       - python-dev
 
 before_install:
-  - PATH="/usr/lib/jvm/java-8-openjdk-amd64/jre/bin":$PATH
-  - JAVA_HOME=/usr/lib/jvm/java-8-openjdk-amd64
+  - PATH="/usr/lib/jvm/java-11-openjdk-amd64/jre/bin":$PATH
+  - JAVA_HOME=/usr/lib/jvm/java-11-openjdk-amd64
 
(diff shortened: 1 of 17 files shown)


wisechengyi:
Just a tip: in case more frequent travis iteration is needed, you could link travis to your own fork, so you'd be able to restart builds and what not.

tpasternak:
`FilePathRelativeToBuiltRootMacroTest#testFilePathRelativeMacro` covered here https://github.com/pantsbuild/intellij-pants-plugin/pull/499

marek1840:
The only remaining error is:
```
File "/opt/python/3.7.1/lib/python3.7/subprocess.py", line 1516, in _execute_child
  raise child_exception_type(errno_num, err_msg, err_filename)                    

  Exception message: [Errno 2] No such file or directory: '/usr/local/lib/jvm/openjdk11/bin/wsimport': '/usr/local/lib/jvm/openjdk11/bin/wsimport'
```
which suggests that the changes made to pants repository in https://github.com/pantsbuild/pants/pull/9216 weren't picked up. Can it be that caches got stale? @wisechengyi could you restart the CI with caches cleared?

PR #499, opened 2020-02-27
[IDEA 201.x] Remove macro registration hack

tpasternak:
In old versions of IntelliJ platform it was only possible to register
macros of custom names. The only way to provide custom macro was to
override one of the existing macros. We've been overriding
`FileRelativePath` macro.

Unfortunately, since version 201.x of IntelliJ platform the hack does
not work, because the data structure to keep the macros has been
changed from a HashMap to a List. This means, the entries are not
overriden, but appended to the end of the list. MacroManager looks up
only for the first ocurrence of the name, so the one in the end of the
list will never be found if the name is not unique. The change was
done in this commit https://github.com/JetBrains/intellij-community/commit/61838067fdd286140a51df64c9fcdbebc1543786

Luckily, the hack is not needed since
https://github.com/JetBrains/intellij-community/commit/fdf6dbcb8ccace8d9cad4f7e6400df834e197687.
This means, we can use any custom name for the macro

### DIFF
--- a/src/com/twitter/intellij/pants/macro/FilePathRelativeToBuiltRootMacro.java
+++ b/src/com/twitter/intellij/pants/macro/FilePathRelativeToBuiltRootMacro.java
@@ -3,7 +3,6 @@
 
 package com.twitter.intellij.pants.macro;
 
-import com.intellij.application.options.PathMacrosImpl;
 import com.intellij.ide.macro.Macro;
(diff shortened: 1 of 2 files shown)


wisechengyi:
Weird that travis link didn't show up here.

https://travis-ci.com/pantsbuild/intellij-pants-plugin/builds/150842760

PR #501, opened 2020-03-05
Support for Idea 201.6073.9

marek1840:


common/com/twitter/intellij/pants/util/PantsUtil.java
```diff
@@ -979,4 +980,15 @@ public static void invalidatePluginCaches() {
   public static List<String> parseCmdParameters(Optional<String> cmdArgsLine) {
     return cmdArgsLine.map(ParametersListUtil::parse).orElse(ContainerUtil.newArrayList());
   }
+
+  public static void invokeLater(Runnable task) {
```
tpasternak: Maybe let's rename it if it's not always "later"

marek1840: Renamed to `invokeLaterIfNeeded` to maintain compatibility with the underlying idea method

src/com/twitter/intellij/pants/components/impl/PantsActionConfigurationCustomizer.java
```diff
@@ -0,0 +1,53 @@
...
+
+    AnAction pantsRebuildAction = new PantsOverrideAction(
+      PantsConstants.REBUILD_PROJECT_DESCRIPTION,
+      new PantsRebuildAction(),
+      manager.getAction(PANTS_REBUILD_ACTION_NAME)
+    );
+
+    manager.registerAction(PANTS_COMPILE_PROJECT_ACTION_NAME, pantsCompileAllTargetAction);
+    manager.registerAction(PANTS_COMPILE_MODULE_ACTION_NAME, pantsMakeModuleAction);
+    manager.registerAction(PANTS_REBUILD_ACTION_NAME, pantsRebuildAction);
```
tpasternak: Can we register actions in `plugin.xml`?

tpasternak: Nvm, it was moved from the other place, it's not a part of this PR

olafurpg:
Not sure if it's relevant here, but got a message from the Scala plugin team that there have been some internal breaking changes 

> If you have plugins for IDEA that depend on Scala plugin compiler or somehow rely on it's behaviour, please check that all is working in 2020.1 EAP. There wer some changes in that area

marek1840:
> there have been some internal breaking changes in Scala plugin

I haven't noticed anything unusual when working on this update, so let's hope those don't affect us here.

PR #502, opened 2020-03-10
Support for refreshing bsp projects on BUILD file change

lukaszwawrzyk:
There is already a file watcher that creates a popup notification with link to project refresh when BUILD file changes.
This PR updates the method that the hyperlink points to - `PantsUtil.refreshAllProjects`, so that in case of project imported from bsp rather than from pants, it would also refresh the project correctly.

Also the change in findBuildRoot was needed (look for pants root in module content roots if it was not found from .iml file) as otherwise the root module could not be found given layout generated by fastpass. Because of that the file watcher was never setup.


wisechengyi:
Thanks, @lukaszwawrzyk !

PR #504, opened 2020-03-12
Save incremental import depth setting

tpasternak:
This change was reverted, because we thought it broke CI. But after
reversion the CI is still red with the same errors.

ORIGINAL COMMIT HASH
1f72954

ORIGINAL MESSAGE

Before this commit, IntelliJ plugin was asking for incremental import
depth each time, even when doing refresh. The depth value was not
saved anywhere.

Now, the incremental import depth parameter is saved alongside other
settings, and may be configured in the same place as others.


tpasternak:
Please don't merge before https://github.com/pantsbuild/intellij-pants-plugin/pull/501 - this will probably fix the CI build

olafurpg:
@tpasternak is this PR ready for merge now?

tpasternak:
Yes, I only need to rebase it

PR #508, opened 2020-03-20
Use isSynthetic over isCodegen

wisechengyi:
Context:
Pants export deprecated `is_code_gen` a while ago and need to use `is_synthetic` instead.

### DIFF
--- a/common/com/twitter/intellij/pants/model/TargetAddressInfo.java
+++ b/common/com/twitter/intellij/pants/model/TargetAddressInfo.java
@@ -28,10 +28,10 @@ public class TargetAddressInfo {
    */
   private String pants_target_type = null;
 
-  private boolean is_code_gen;
+  private boolean is_synthetic;
 
-  public boolean isCodeGen() {
-    return is_code_gen;
+  public boolean isSynthetic() {
+    return is_synthetic;
   }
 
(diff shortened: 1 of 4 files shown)


common/com/twitter/intellij/pants/model/TargetAddressInfo.java
```diff
@@ -28,10 +28,10 @@
    */
   private String pants_target_type = null;
 
-  private boolean is_code_gen;
+  private boolean is_synthetic;
 
-  public boolean isCodeGen() {
-    return is_code_gen;
+  public boolean isSynthetic() {
+    return is_synthetic;
```
cosmicexplorer: Does this generate a json parser? Where are we changing our JSON parsing to consume this new field?

wisechengyi: A gson parser is using certain conventions: https://github.com/pantsbuild/intellij-pants-plugin/blob/c3ae7eaf7f75ca2fea1194ccb9b78635b607974a/src/com/twitter/intellij/pants/service/project/model/TargetInfoDeserializer.java#L36 to variable name to the json field.

cosmicexplorer: Awesome, thank you!

PR #515, opened 2020-04-07
Fix debugger not starting for junit tests

marek1840:


marek1840:
Note, that the new pants master introduced a failure in `com.twitter.intellij.pants.integration.OSSPantsCompileActionsTest#testLintTargetAction`. It will be handled in a separate PR and after it gets merged, I will rebase this one.

marek1840:
> Note, that the new pants master introduced a failure in com.twitter.intellij.pants.integration.OSSPantsCompileActionsTest#testLintTargetAction

And yet, it didn't fail on the CI. I have confirmed with @tpasternak that this error **did not** happen on his machine, so I think it is best to assume that the problem is somewhere on my local machine and the PR can be considered merge-ready.

olafurpg:
Do we need a new pants plugin release to get this distributed via MDE?

PR #516, opened 2020-04-17
Action to convert pants project to bsp

lukaszwawrzyk:
Adds an action to Pants menu in Pants projects that starts a background task. This task runs fastpass to generate project with the same name as the original project has and the same targets as were selected for the project.

### DIFF
--- a/resources/META-INF/plugin.xml
+++ b/resources/META-INF/plugin.xml
@@ -73,6 +73,10 @@
               class="com.twitter.intellij.pants.compiler.actions.PantsCompileCurrentTargetAction"
               text="Compile target(s) in the selected editor">
       </action>
+      <action id="com.twitter.intellij.pants.ui.PantsToBspProjectAction"
+              class="com.twitter.intellij.pants.ui.PantsToBspProjectAction"
+              text="Convert to BSP project">
+      </action>
     </group>
 
     <group id="Pants.Compile" text="Pants Compile" description="Pants compilation options" icon="PantsIcons.Icon"
--- a/src/com/twitter/intellij/pants/ui/PantsToBspProjectAction.java
+++ b/src/com/twitter/intellij/pants/ui/PantsToBspProjectAction.java
@@ -1,0 +1,130 @@
+// Copyright 2020 Pants project contributors (see CONTRIBUTORS.md).
+// Licensed under the Apache License, Version 2.0 (see LICENSE).
+
+package com.twitter.intellij.pants.ui;
+
+import com.intellij.execution.configurations.GeneralCommandLine;
+import com.intellij.ide.impl.OpenProjectTask;
+import com.intellij.ide.impl.ProjectUtil;
+import com.intellij.openapi.actionSystem.AnAction;
+import com.intellij.openapi.actionSystem.AnActionEvent;
+import com.intellij.openapi.application.ApplicationManager;
+import com.intellij.openapi.progress.ProgressIndicator;
+import com.intellij.openapi.progress.ProgressManager;
+import com.intellij.openapi.progress.Task;
+import com.intellij.openapi.project.DumbAware;
+import com.intellij.openapi.project.Project;
+import com.intellij.openapi.ui.Messages;
+import com.twitter.intellij.pants.settings.PantsSettings;
+import com.twitter.intellij.pants.util.PantsUtil;
+import org.apache.commons.io.IOUtils;
+import org.jetbrains.annotations.NotNull;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.net.URL;
+import java.nio.charset.StandardCharsets;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.Arrays;
+import java.util.List;
+import java.util.stream.Collectors;
+
+public class PantsToBspProjectAction extends AnAction implements DumbAware {
+  private static final String METALS_VERSION = "0.8.4";
+
+  @Override
+  public void update(@NotNull AnActionEvent e) {
+    Project project = e.getProject();
+    boolean isPants = project != null && PantsUtil.isPantsProject(project);
+    e.getPresentation().setEnabledAndVisible(isPants);
+  }
+
+  @Override
+  public void actionPerformed(AnActionEvent e) {
+    Project project = e.getProject();
+    if (project == null) {
+      Messages.showInfoMessage("Project not found.", "Error");
+      return;
+    }
+
+    ProgressManager.getInstance().run(new Task.Backgroundable(project, "Preparing BSP Project", false) {
+      @Override
+      public void run(@NotNull ProgressIndicator indicator) {
+        try {
+          GeneralCommandLine commandLine = createCommandLine(project);
+          Process process = commandLine.createProcess();
+          String output = readToString(process.getInputStream());
+          String outputErr = readToString(process.getErrorStream());
+          int result = process.waitFor();
+          if (result != 0) {
+            String message = formatError(output, outputErr);
+            throw new RuntimeException(message);
+          } else {
+            Path projectPath = Paths.get(output);
+            ProjectUtil.openOrImport(projectPath, new OpenProjectTask());
+          }
+        }
+        catch (Exception ex) {
+          ApplicationManager.getApplication()
+            .invokeLater(() -> Messages.showErrorDialog(project, ex.getMessage(), "Error"));
+        }
+      }
+    });
+  }
+
+  private GeneralCommandLine createCommandLine(Project project) throws IOException {
+    GeneralCommandLine commandLine = PantsUtil.defaultCommandLine(project);
+
+    String coursier = coursierPath().toString();
+    commandLine.setExePath(coursier);
+
+    String name = project.getName();
+    List<String> commandBase = Arrays.asList(
+      "launch", "org.scalameta:metals_2.12:" + METALS_VERSION,
+      "--main", "scala.meta.internal.pantsbuild.BloopPants",
+      "--",
+      "create",
+      "--name", name,
+      "--intellij",
+      "--intellijLauncher", "echo"
+    );
+    commandLine.addParameters(commandBase);
+
+    List<String> targets = pantsTargets(project);
+    commandLine.addParameters(targets);
+    return commandLine;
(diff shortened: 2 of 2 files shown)


wisechengyi:
Not sure if the conversion succeeded the first time, but this is the second time I ran it. 

1) Possibly need a check that if it's already a BSP project, then the option should be greyed out.
2) We could use a UUID for the new project path for the converted project. It could also be useful if user initially imported with `./pants idea-plugin <long list of targets>` so the filename won't be longer than what the OS can handle.

Screen Shot 2020-04-17 at 10 03 10 PM

lukaszwawrzyk:
@olafurpg 
Now the project name is just reused from the original project.
Generally there are some problems here:
If I delegate naming and location to fastpass, I can only know the path of the project when it is generated, then I would need to store it somewhere in this pants project in order to know that project was already created. With this I could just check if directory exists and open the project without running fastpass amend or create. I could then change action's presentation to say "Open connected BSP project" or something, or just hide it based on existence of that dir.

Running amend in case create fails can be a waste of time as amending might not be necessary at all, we just want to open the project. To run amend we need to know at least the project name, so we need some deterministic way to compute it from current project, and can't delelegate naming to fastpass.

@wisechengyi It should have prompted you to open a new project, on first run and on second run fail with this error. 
So another approach is with this UUID. The problem is probably user experience and difficulty to find which project is which from dir name. So another idea from this is that we each time still create a new project if this action is run, with partially random name and we would never fail to convert, but not sure if we want it.

I think that we can get the best UX with storing linked bsp project and use that info (with naming delegated to fastpass).

resources/META-INF/plugin.xml
```diff
@@ -73,6 +73,10 @@
               class="com.twitter.intellij.pants.compiler.actions.PantsCompileCurrentTargetAction"
               text="Compile target(s) in the selected editor">
       </action>
+      <action id="com.twitter.intellij.pants.ui.PantsToBspProjectAction"
+              class="com.twitter.intellij.pants.ui.PantsToBspProjectAction"
+              text="Convert to BSP project">
```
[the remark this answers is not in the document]
lukaszwawrzyk: I did it with the longer one, I think we have space for it :)

src/com/twitter/intellij/pants/ui/PantsToBspProjectAction.java
```diff
@@ -0,0 +1,130 @@
...
+import java.nio.charset.StandardCharsets;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.Arrays;
+import java.util.List;
+import java.util.stream.Collectors;
+
+public class PantsToBspProjectAction extends AnAction implements DumbAware {
+  private static final String METALS_VERSION = "0.8.4";
```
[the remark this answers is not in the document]
lukaszwawrzyk: I expected coursier to have something like this, but didn't find it initially. Changed.

olafurpg:
@lukaszwawrzyk That makes sense! I agree it's desirable that the Pants plugin can compute the fastpass name to test whether a project already exists or not.

In this case we can try to redo some of the logic in fastpass around naming the project

https://github.com/scalameta/metals/blob/9d6f2e99e1b09904b38b72393bd8b123677041dc/metals/src/main/scala/scala/meta/internal/pantsbuild/PantsConfiguration.scala#L114-L128

It's important that the project name is to too long and that it doesn't contain special characters (especially not `:`).

lukaszwawrzyk:
@olafurpg So I already pushed an implementation that relies on storing project path that we get when we create the project with fastpass. WIth this we can easily open it if it exists and we don't need to duplicate naming logic or call fastpass amend.

If we were to generate names from pants plugin and based on that check if project exists, we would also need to know the parent directory where these bsp projects should be generated (so also copy this logic). So what I currently have, saves us from reimplementing anything from fastpass. Unless there is a reason it's not desirable to store the linked bsp project path within pants project.

src/com/twitter/intellij/pants/ui/PantsToBspProjectAction.java
```diff
@@ -0,0 +1,165 @@
...
+      "launch", "org.scalameta:metals_2.12:latest.stable",
+      "--main", "scala.meta.internal.pantsbuild.BloopPants",
+      "--",
+      "create",
+      "--intellij",
+      "--intellijLauncher", "echo"
+    );
+    commandLine.addParameters(commandBase);
+
+    List<String> targets = pantsTargets(project);
```
olafurpg, line 113: It looks like these are absolute paths since the generated project is named `.Users.lgeirsson.workspace.source...`

Screenshot 2020-04-21 at 10 53 17


I think we can fix this by stripping the prefix containing the absolute path to the Pants workspace away
```diff
- INFO  process: /Users/lgeirsson/workspace/source/pants ... /Users/lgeirsson/workspace/source/buildcache::
+ INFO  process: /Users/lgeirsson/workspace/source/pants ... buildcache::
```

src/com/twitter/intellij/pants/ui/PantsToBspProjectAction.java
```diff
@@ -0,0 +1,165 @@
...
+
+  private GeneralCommandLine createCommandLine(Project project) throws IOException {
+    GeneralCommandLine commandLine = PantsUtil.defaultCommandLine(project);
+
+    String coursier = coursierPath().toString();
+    commandLine.setExePath(coursier);
+
+    //String name = project.getName();
+    List<String> commandBase = Arrays.asList(
+      "launch", "org.scalameta:metals_2.12:latest.stable",
```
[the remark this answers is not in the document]
wiwa: We need to be able to have a `fastpass.properties` config file on the classpath as implemented in https://github.com/scalameta/metals/pull/1582/files#diff-1ff86c347afd6af0342fc1157784b735R39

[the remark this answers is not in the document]
olafurpg: @lukaszwawrzyk please ignore this comment, we figured out a way to solve this problem so that this coursier launch command works unchanged.

src/com/twitter/intellij/pants/ui/PantsToBspProjectAction.java
lukaszwawrzyk: @olafurpg I will push potential fix, Yet I couldn't reproduce. If you could tell me how you setup the pants project to get this problem I could test it.
I just take bsp-pants repo on branch scratchfile, I do import pants project pointing to 'scratchfile' directory and proceed with defaults. This results in targets being a `List("scratchfile/::")` and project is named "scratchfile.".

src/com/twitter/intellij/pants/ui/PantsToBspProjectAction.java
```diff
@@ -126,13 +127,19 @@ private Path coursierPath() throws IOException {
...
     PantsSettings pantsSettings = PantsSettings.getInstance(project);
     return pantsSettings.getLinkedProjectsSettings()
       .stream()
       .flatMap(projectSettings -> projectSettings.getSelectedTargetSpecs().stream())
+      .map(target -> stripPrefix(target, root))
       .collect(Collectors.toList());
   }
 
+  private String stripPrefix(String s, String prefix) {
+    return s.startsWith(prefix) ? s.substring(prefix.length()) : s;
```
olafurpg, line 140: This leaves a leading `/` in the start resulting in project names like `.foo`

lukaszwawrzyk: @olafurpg I implemented removing the `"/"` Hope it works now, as I don't have reproduction for full directory in target names.

src/com/twitter/intellij/pants/ui/PantsToBspProjectAction.java
olafurpg: That fixed the issue. Thanks!

src/com/twitter/intellij/pants/ui/PantsToBspProjectAction.java
```diff
@@ -0,0 +1,192 @@
...
+
+    List<String> targets = pantsTargets(project);
+    commandLine.addParameters(targets);
+    return commandLine;
+  }
+
+  private Path coursierPath() throws IOException {
+    Path destination = Paths.get(System.getProperty("java.io.tmpdir"), "pants-plugin-coursier");
+    if (!Files.exists(destination)) {
+      URL url = new URL("https://git.io/coursier-cli");
```
tpasternak, line 142: We download newest coursier here. Imo this should either be changed to  some fixed version, or logged. Without it we lose information about coursier version that may be essential to reproduce potential bugs.

src/com/twitter/intellij/pants/ui/PantsToBspProjectAction.java
```diff
@@ -0,0 +1,192 @@
...
+    commandLine.setExePath(coursier);
+
+    List<String> commandBase = Arrays.asList(
+      "launch", "org.scalameta:metals_2.12:latest.stable",
+      "--main", "scala.meta.internal.pantsbuild.BloopPants",
+      "--",
+      "create",
+      "--intellij",
+      "--intellijLauncher", "echo"
+    );
```
tpasternak, line 131: It would be nice to log downloaded metals version or fix it in code.

src/com/twitter/intellij/pants/ui/PantsToBspProjectAction.java
olafurpg: This is a good point. If we copy the coursier bootstrap script from resources then we always use the same coursier version

src/com/twitter/intellij/pants/ui/PantsToBspProjectAction.java
olafurpg: I don't think we can tell the exact Metals version from this command when using `latest.stable`. One alternative would be to additionally launch a `version` command so that we can log what version got picked up 🤔

tpasternak: Right, but I think it's not essential for this PR, so it can be done as a follow-up (as well as the cousier-verison related issue). Nevertheless it must be done soon, as bug fixing may be very hard without it, in some specific cases.

PR #518, opened 2020-04-21
Shorten target names in bsp library origin names

lukaszwawrzyk:
Depends on
https://github.com/JetBrains/intellij-scala/pull/529

### DIFF
--- a/resources/META-INF/pants-scala.xml
+++ b/resources/META-INF/pants-scala.xml
@@ -13,6 +13,7 @@
   <extensions defaultExtensionNs="com.intellij">
     <externalProjectDataService implementation="com.twitter.intellij.pants.service.scala.PantsScalaDataService" order="last"/>
     <highlightVisitor implementation="com.twitter.intellij.pants.highlight.PantsScalaHighlightVisitor"/>
+    <bspResolverNamingExtension implementation="com.twitter.intellij.pants.service.scala.PantsBspResolverNamingExtension"/>
   </extensions>
   <extensions defaultExtensionNs="com.intellij.plugins.pants">
     <projectResolver implementation="com.twitter.intellij.pants.service.scala.ScalaSdkResolver"/>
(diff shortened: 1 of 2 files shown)


olafurpg:
@lukaszwawrzyk is this PR ready for merge?

lukaszwawrzyk:
Not sure if/how I need to bump scala plugin here. I retriggered CI to see if it will build now @olafurpg

olafurpg:
@wisechengyi do you know how the Scala plugin version is selected in this build?

wisechengyi:
it auto picks the version associated with an intellij build number: https://github.com/pantsbuild/intellij-pants-plugin/blob/master/scripts/setup-ci-environment.sh#L56

lukaszwawrzyk:
@wisechengyi So it seems like it picks some older version. Shall we wait for release or just hardcode latest version for scala plugin?

wisechengyi:
Which version of scala plugin does it require? I checked today it doesn't work with 2019.1.29 yet.

I don't mind consuming a nightly build of scala plugin here, but chances we won't release a nightly version internally, so it might be okay to wait, as the latency for us to consume changes in the repo is fairly low.

lukaszwawrzyk:
> Which version of scala plugin does it require? I checked today it doesn't work with 2019.1.29 yet.
> 
> I don't mind consuming a nightly build of scala plugin here, but chances we won't release a nightly version internally, so it might be okay to wait, as the latency for us to consume changes in the repo is fairly low.

If we would run this version of pants plugin (after merge) with old scala plugin it wouldn't work correctly, so if you don't release scala plugin internally it would be better to wait. 
Indeed the change is not in 2020.1.29. I checked 2020.1.1173 nightly and it is there.
I asked Justin on when we will have another stable release.

@

olafurpg:
Sounds good. We can wait with merging this for now. We're preparing a new internal release now and we can include this fix in the next release (which will hopefully happen later this month)

olafurpg:
@wisechengyi any idea how we can update the version of the Scala plugin in the Gradle build to make this PR green?

wisechengyi:
It will happen automatically when jetbrains releases the latest stable version of scala plugin.

olafurpg:
Sounds good, let's wait then 👍

olafurpg:
The Scala plugin had a stable release yesterday (May 19th), see 2020.1.36 https://plugins.jetbrains.com/plugin/1347-scala/versions

lukaszwawrzyk:
Pushed to trigger build again

olafurpg:
The CI is green \o/

PR #519, opened 2020-04-21
Always set "JAR manifest" cmd line shortener for new run configs

tpasternak:


src/com/twitter/intellij/pants/execution/SetJarManifestCommandLineShortener.scala
```diff
@@ -0,0 +1,15 @@
...
+import com.intellij.execution.{ConfigurationWithCommandLineShortener, RunManagerListener, RunnerAndConfigurationSettings, ShortenCommandLine}
+
+class SetJarManifestCommandLineShortener extends RunManagerListener {
+  override def runConfigurationAdded( settings: RunnerAndConfigurationSettings) {
+    settings.getConfiguration match {
+      case configuration: ConfigurationWithCommandLineShortener => configuration.setShortenCommandLine(ShortenCommandLine.MANIFEST)
+      case _ =>
+    }
+  }
+}
```
wisechengyi, line 15: Given pants plugin doesn't bring scala runtime into intellij but relies only on intellij's scala plugin, so if scala plugin isn't enabled, it will likely explode here. Do you think it would be okay to rewrite this java?

If this is only for running scala main class, we can move the meta info `resources//META-INF/pants-scala.xml` and keep this scala.

tpasternak: Right, thanks!

PR #526, opened 2020-05-04
Add automatic fastpass update notification

lukaszwawrzyk:
After loading first project there is a thread that starts watching the system fastpass version located at `/opt/twitter_mde/bin/fastpass`. If version is different (read from bloop.settings.json) it creates a notification with "Update" link. When clicked it kills bloop, runs `fastpass refresh` command (with the system version) and triggers project refresh.

### DIFF
--- a/src/com/twitter/intellij/pants/components/impl/FastpassUpdater.java
+++ b/src/com/twitter/intellij/pants/components/impl/FastpassUpdater.java
@@ -104,8 +104,8 @@ public class FastpassUpdater {
     }
   }
 
-  public static void initialize() {
-    if (!initialized) {
+  public static void initialize(Project project) {
+    if (PantsUtil.isBspProject(project) && !initialized) {
       synchronized (FastpassUpdater.class) {
         if (!initialized) {
           ScheduledExecutorService timer = Executors.newScheduledThreadPool(1);
--- a/src/com/twitter/intellij/pants/components/impl/FastpassUpdater.java
+++ b/src/com/twitter/intellij/pants/components/impl/FastpassUpdater.java
@@ -38,6 +38,7 @@ import java.util.Optional;
 import java.util.concurrent.Executors;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
+import java.util.function.Supplier;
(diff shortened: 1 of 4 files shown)


lukaszwawrzyk:
I updated the PR. Now it is checked once a day, additionally there is an action in Pants menu (visible only if fastpass is available under the that twitter_mde path).

src/com/twitter/intellij/pants/components/impl/FastpassUpdater.java
```diff
@@ -58,18 +61,59 @@
...
+      VirtualFile fastpassBinary = LocalFileSystem.getInstance().findFileByPath(FASTPASS_PATH);
+      return fastpassBinary != null && fastpassBinary.exists();
+    }
+  }
+
   public static void initialize() {
     if (!initialized) {
       synchronized (FastpassUpdater.class) {
         if (!initialized) {
           ScheduledExecutorService timer = Executors.newScheduledThreadPool(1);
```
olafurpg, line 111: Just to confirm, does this check happen for users who are not using a BSP project?

lukaszwawrzyk: Good catch, I lost this logic somewhere on the way. So it would always start the thread, but not do anything than check for open bsp projects once a day. I updated it, so that the thread is started after opening first bsp project (rather than any project).

src/com/twitter/intellij/pants/components/impl/FastpassUpdater.java
```diff
@@ -0,0 +1,254 @@
...
+      refresh.waitFor();
+      return refresh.exitValue() == 0;
+    }
+    catch (Exception e) {
+      LOG.warn(e);
+      return false;
+    }
+  }
+
+  private static Optional<FastpassData> extractFastpassData(Project project) {
```
olafurpg, line 209: I just noticed that this method relies on the implementation details of `bloop.settings.json` that may change in the future. Can we first add a check that tries to read the `fastpassVersion` field in `.bsp/bloop.json`? This field is currently not generated by fastpass so it won't be used now, but we can add it separately to Metals which would make this logic less fragile.

PR #528, opened 2020-05-08
Add GUI for `faspass amend` command

tpasternak:
Allows user to include (or remove) new pants targets into already
imported project


tpasternak:
There are some opportunities to improve user experience which maybe achieved with relatively small effort:
1. Auto-completion
2. Prefetching targets into cache while user is typing.

tpasternak:
Labels changed and small parsing improvement added

olafurpg:
I just tried this again and it's working well :+1: 

If it's easy to implement, I'm wondering if we could somehow warn the user about pressing "OK" to save the amendment when there are 0 matching Pants targets? I don't think we should prevent the user from pressing "OK" in this case, but there's currently no strong visual hint (for example color or warning message) that pressing OK will most likely cause the refresh to fail.

olafurpg:
Otherwise, LGTM!

tpasternak:
Added such a warning. Please let me know if it's enough. We can add "Are you sure?" message box as well.

image

olafurpg:
Is the CI failure expected?

tpasternak:
No, it's not, I pushed a commit to rerun

olafurpg:
@wisechengyi any chance we could merge this change?

PR #534, opened 2020-05-27
Use new artifact for fastpass

Duhemm:
Fastpass used to be part of the Metals repository, but it was recently
moved to its own repository. It is no longer published as part of
metals.

### DIFF
--- a/src/com/twitter/intellij/pants/bsp/FastpassUtils.java
+++ b/src/com/twitter/intellij/pants/bsp/FastpassUtils.java
@@ -98,8 +98,8 @@ final public class FastpassUtils {
 
 
   private static List<String> coursierPart(){
-    return Arrays.asList("launch", "org.scalameta:metals_2.12:latest.stable",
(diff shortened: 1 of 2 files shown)


src/com/twitter/intellij/pants/bsp/FastpassUtils.java
```diff
@@ -98,8 +98,8 @@
 
 
   private static List<String> coursierPart(){
-    return Arrays.asList("launch", "org.scalameta:metals_2.12:latest.stable",
-                         "--main", "scala.meta.internal.pantsbuild.BloopPants",
+    return Arrays.asList("launch", "org.scalameta:fastpass_2.12:latest.stable",
+                         "--main", "scala.meta.internal.fastpass.pantsbuild.BloopPants",
```
olafurpg, line 102: I'm preparing a PR to add this main class here
```suggestion
                         "--main", "scala.meta.fastpass.Fastpass",
```

PR #538, opened 2020-06-05
Bump to IJ 202.6109.22

tpasternak:


tpasternak:
Waiting only for this https://github.com/JetBrains/intellij-scala/pull/545

tpasternak:
Please don't merge yet, the newest EAP introduced a lot of new parameters to OpenProjectTask constructor, I'm not sure yet if they're valid

tpasternak:
Finaly the CI is green. The last force-push only squashes a few commits

testFramework/com/twitter/intellij/pants/testFramework/PantsIntegrationTestCase.java
```diff
@@ -517,6 +520,7 @@ public void tearDown() throws Exception {
       myCompilerTester.tearDown();
     }
 
+
```
wisechengyi, line 523: xx

resources/META-INF/plugin.xml
```diff
@@ -16,7 +16,7 @@
   <version>1.15.1</version>
   <vendor>Twitter, Inc.</vendor>
 
-  <idea-version since-build="201.6073" until-build="201.*"/>
+  <idea-version since-build="201.6073" until-build="202.*"/>
```
wisechengyi, line 19: just checking that there's no breaking API changes from 201.x right? otherwise we'll need to be bump `since-build`?

tpasternak: the `OpenProjectTask` signature has changed, so indeed, it would be safer to bump the `since-build` field

tpasternak:
@wisechengyi I force-pushed a new version with recommendations applied

resources/META-INF/plugin.xml
wisechengyi: ty!

olafurpg:
The CI is green now, can we merge this PR? @wisechengyi

PR #540, opened 2020-06-18
Enable ideprobe for pants plugin CI

lukaszwawrzyk:
This PR enables a way to run ide-probe tests for intellij pants plugin.

I added new element to travis matrix (ideprobe) which runs test.sh script inside docker. 

The docker image just setups packages needed to run probe and pants (python, xvfb etc.) and coursier and accepts an agreement needed to run intellij. 

The `run_docker` script builds pants plugin and passes it down to docker container.

test.sh script is used inside docker to run tests. It enables virtual display to run tests with full gui. Then it installs ideprobe artifacts to ivy local repo (eventually we should just use a published version and proper resolvers). 

Tests are run using junit standalone launcher (run via coursier). Coursier is also used to build the classpath required to run ide-probe (the framework and tests). The tests for pants are inside artifact 
`com.virtuslab.ideprobe:pants-tests_2.13:0.1:test` and the jar for this artifact is passed as `scan-classpath` flag for junit to run only pants related tests. 

There is one more thing here. At the beginning of runtime classpath we put one more jar with `pants.conf` file. In ide-probe it contains common config for pants tests: intellij and plugin versions. Thanks to this we control have control over versions here and also we can use the pants plugin from current PR. In similar way it is possible to override any other probe config externally.

Further things to do with this integration:
- Note that I disabled PantsIssue42 test, it is quite heavy as it reproduces an issue and ensures it's not there anymore, but to do this is sets up 2 extra versions of pants. Running this causes exceed of the max 90min time for a travis job. In general, tests will need to be split (probably with set of more granular include/exclude patterns to fit in. 
- Screenshots - ide probe takes screenshots during ui tests, we can make them available to look at for failing tests. We will just need to see how to do this in travis and implement it.


ideprobe/Dockerfile
```diff
@@ -0,0 +1,13 @@
+FROM openjdk:11
```
wisechengyi, line 1: Note that the plugin built by Pants is still via JDK 8, in case any discrepancy occurs in the future.

PR #541, opened 2020-06-24
Create synthetic root modules modules for target specs

tpasternak:
Thanks to this we decrease the number of separate module graphs, which hopefully improves performance of VCS-related IntelliJ features.

### DIFF
--- a/src/com/twitter/intellij/pants/service/project/resolver/PantsSourceRootsExtension.java
+++ b/src/com/twitter/intellij/pants/service/project/resolver/PantsSourceRootsExtension.java
@@ -8,6 +8,7 @@ import com.intellij.openapi.externalSystem.model.ProjectKeys;
 import com.intellij.openapi.externalSystem.model.project.ContentRootData;
 import com.intellij.openapi.externalSystem.model.project.ModuleData;
 import com.intellij.openapi.externalSystem.model.project.ProjectData;
+import com.intellij.openapi.module.ModuleTypeId;
 import com.intellij.openapi.util.io.FileUtil;
 import com.intellij.openapi.util.text.StringUtil;
 import com.intellij.util.containers.ContainerUtil;
@@ -19,15 +20,24 @@ import com.twitter.intellij.pants.service.project.model.ProjectInfo;
 import com.twitter.intellij.pants.service.project.model.TargetInfo;
 import com.twitter.intellij.pants.service.project.model.graph.BuildGraph;
 import com.twitter.intellij.pants.util.PantsConstants;
+import org.apache.commons.compress.utils.Lists;
 import org.jetbrains.annotations.NotNull;
 
+import java.io.File;
+import java.nio.file.Path;
+import java.nio.file.Paths;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collection;
+import java.util.Collections;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
 import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 public class PantsSourceRootsExtension implements PantsResolverExtension {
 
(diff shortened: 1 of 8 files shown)


src/com/twitter/intellij/pants/service/project/resolver/PantsSourceRootsExtension.java
```diff
@@ -67,6 +77,68 @@ public void resolve(
...
+        .flatMap(x -> x.getChildren()
+        .stream())
+        .map(x -> x.getData(ProjectKeys.CONTENT_ROOT))
+        .filter(Objects::nonNull)
+        .map(x -> x.getRootPath())
+        .collect(Collectors.toSet());
+    replacedWithoutChildren.removeAll(allContentRoots);
+
+    for (String path: replaced) {
+      String moduleName = projectPath.relativize(Paths.get(path)).toString().replaceAll("/", "_") + "-project-root";
```
olafurpg, line 126: Fastpass uses the `-project-root` naming convention but it still feels a bit weird. Any suggestions for a clearer or more self-explanatory name? 🤔

src/com/twitter/intellij/pants/service/project/resolver/PantsSourceRootsExtension.java
```diff
@@ -67,6 +77,68 @@ public void resolve(
...
+      final ModuleData moduleData = new ModuleData(
```
olafurpg, line 127: (unrelated to this PR) This module doesn't have any content roots from what I can see. Is there a way to achieve exactly the same via BSP? When I tried to create empty Bloop projects then it didn't help, I had to add a fake resource directory which caused other problems.

src/com/twitter/intellij/pants/service/project/resolver/PantsSourceRootsExtension.java
tpasternak: I'm not sure. It could be something less commonly used in build systems, so  it's easier to search it it code/issue tracker. So for example "common ancestor"

src/com/twitter/intellij/pants/service/project/resolver/PantsSourceRootsExtension.java
tpasternak: the content root is added below

src/com/twitter/intellij/pants/service/project/resolver/PantsSourceRootsExtension.java
```diff
@@ -67,6 +77,72 @@ public void resolve(
...
+
+  private void installSyntheticModules(
+    @NotNull PantsCompileOptionsExecutor executor,
+    @NotNull DataNode<ProjectData> projectDataNode,
+    @NotNull Map<String, DataNode<ModuleData>> modules
+  ) {
+    Path projectPath = Paths.get(executor.getBuildRoot().getAbsolutePath());
+    Set<String> targetSpecPaths =
+      executor.getOptions().getSelectedTargetSpecs().stream()
+        .map(x -> x.replaceFirst("/?:.*", ""))
```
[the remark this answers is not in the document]
tpasternak: I copied the regex from here, it was assumed that the spec may be of the form `spec_path/:spec_target` I'm not sure if it's possible at all, but kept it just in case

https://github.com/scalameta/fastpass/blob/0e8092e8e650608aa879427e834cb1b1ab469501/fastpass/src/main/scala/scala/meta/internal/fastpass/pantsbuild/PantsConfiguration.scala#L93-L110

src/com/twitter/intellij/pants/service/project/resolver/PantsSourceRootsExtension.java
```diff
@@ -67,6 +77,72 @@ public void resolve(
...
+      modules.values().stream()
+        .flatMap(x -> x.getChildren()
+        .stream())
+        .map(x -> x.getData(ProjectKeys.CONTENT_ROOT))
+        .filter(Objects::nonNull)
+        .map(x -> x.getRootPath())
+        .collect(Collectors.toSet());
+
+    // todo - optimize from m*n to m * log(n)
+    targetSpecPathsWithoutChildren.removeIf(targetSpec -> allContentRoots.stream().anyMatch(contentRoot ->  projectPath.resolve(targetSpec).startsWith(contentRoot)));
```
tpasternak, line 126: Left a "todo" here because I'm not sure if it's worth doing it. I iterate over the list of all already created content roots for each target spec passed as an argument. This means N*M complexity. It could be optimized to N log(M) if I used trie data structure.

src/com/twitter/intellij/pants/service/project/resolver/PantsSourceRootsExtension.java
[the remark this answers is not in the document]
wisechengyi: Sounds good. Thanks!

PR #545, opened 2020-07-10
Ideprobe 2020.2

lukaszwawrzyk:
This PR bumps ideprobe tests to use Intellij 2020.2.

Also it has a rewrite/reorganization of tests to avoid unnecessary starting intellij so many times for basic open project checks, as well as removes duplicate tests and adds pants/bsp variants for everything that makes sense.

Finally, it has reorganized CI running mechanisms (splitting into separate jobs because tests are quite slow on CI (locally 30 min to complete all) and using a published docker image to shorten time to setup everything.


ideprobe/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM openjdk:11
+FROM openjdk:8
```
tpasternak, line 1: Isn't it better to keep 11?

lukaszwawrzyk: Pants plugin is built on java 8 so I just want to make it safer in terms of compatibility

tpasternak: @marek1840 didn't we need to switch to 11 since 2020.1?

marek1840: If I recall correctly ww were using 11 (only in tests) when the Scala plugin was built using 11. Is it no longer the case?

lukaszwawrzyk: Yi mentioned here that it is on java 8 https://github.com/pantsbuild/intellij-pants-plugin/pull/540

ideprobe/tests/src/test/scala/com/twitter/intellij/pants/Suites.scala
```diff
@@ -0,0 +1,23 @@
...
+
+@RunWith(classOf[Suite])
+@SuiteClasses(Array(
+  classOf[OpenProjectTestBsp],
+  classOf[OpenProjectTestPants],
+  classOf[BUILDFilesTest],
+  classOf[PantsSettingsTest],
+  classOf[ThriftIdeaPluginTest]
+))
+class Suite1
```
marek1840, line 15: Could we use more descriptive names?

lukaszwawrzyk: These are only for sake of grouping. There is no logic behind it except for that both take similar time to complete

marek1840: So the grouping is completely arbitrary and those test classes have nothing in common? In that case we could call them batches and comment on how the grouping works, so the next person adding a test does not need to wonder where to include it.

ideprobe/Dockerfile
wisechengyi: Yeah the plugin is built with 8. However the runtime used for IntelliJ is 11. So either version is fine for now. 

It would be good to explore whether we can have it built with 11 tho.

wisechengyi:
merging per https://github.com/pantsbuild/intellij-pants-plugin/pull/552#issuecomment-665297268. we can follow up the minor points in a separate review.

PR #547, opened 2020-07-16
Enable python facet setup

lukaszwawrzyk:
I uncommented the code to bring back the feature that was disabled. I also changed it to be synchronous as if it is async, the map with interpreters it is supposed to fill will be empty, and the code later can't use it. Also moved setting up sdk paths out of write action. 

The changes in test teardown are related to https://youtrack.jetbrains.com/issue/IDEA-246356 bug that I found.


olafurpg:
@wisechengyi any chance we could merge this fix?

PR #548, opened 2020-07-17
Fix `Failed to write settings` error

Duhemm:
Previously, IntelliJ would try to write the settings at the root of the
filesystem, which would be likely to fail. This commit fixes this
behavior by specifying the absolute path where the settings should be
written.


tpasternak:
I think should be done like this https://github.com/tpasternak/intellij-pants-plugin/commit/1c4f107928de3fc963ca8d5e904b108aaa36aaa9

Duhemm:
Thanks a lot @tpasternak ! Superseded by #549

PR #549, opened 2020-07-17
Pass the right path to iml directory on module creation

tpasternak:
Fixes "failed to save settings" error that sometimes occur

### DIFF
--- a/src/com/twitter/intellij/pants/service/project/resolver/PantsSourceRootsExtension.java
+++ b/src/com/twitter/intellij/pants/service/project/resolver/PantsSourceRootsExtension.java
@@ -133,7 +133,7 @@ public class PantsSourceRootsExtension implements PantsResolverExtension {
         PantsConstants.SYSTEM_ID,
         ModuleTypeId.JAVA_MODULE,
         moduleName,
-        targetPath,
+        projectDataNode.getData().getIdeProjectFileDirectoryPath() + "/" + moduleName,
         new File(executor.getBuildRoot(), relativePath).getAbsolutePath()
       );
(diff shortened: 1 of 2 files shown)


src/com/twitter/intellij/pants/service/project/resolver/PantsSourceRootsExtension.java
```diff
@@ -133,7 +133,7 @@ private void installSyntheticModules(
         PantsConstants.SYSTEM_ID,
         ModuleTypeId.JAVA_MODULE,
         moduleName,
-        targetPath,
+        projectDataNode.getData().getIdeProjectFileDirectoryPath() + "/" + moduleName,
```
Duhemm, line 136: Will that be okay with platforms that use a different file separator?

tpasternak: I'm not sure, but we already have three expressions like this https://github.com/pantsbuild/intellij-pants-plugin/search?q=getIdeProjectFileDirectoryPath&unscoped_q=getIdeProjectFileDirectoryPath

tpasternak:
@wisechengyi we've already merged it to https://github.com/pantsbuild/intellij-pants-plugin/tree/release/1.15.1_patches

Can we also merge it to master?

src/com/twitter/intellij/pants/service/project/resolver/PantsSourceRootsExtension.java
wisechengyi: yeah, file.separator is more hygienic, but since we are only support linux and mac, it works for now.

PR #551, opened 2020-07-23
New "Project Files by Target Specs" view

tpasternak:
Add a new tab to the "Project" panel, that may be used to browse
filesystem, but limits visible directories only to the ones that
directly match rules defined in Target Specs, so it does not show
folders containing dependencies unless they are not outside the target
spec directories.


olafurpg:
I just tested this PR and it's amazing! Is there a way for the "Project files by Target Specs" to be selected by default when creating a new project?

tpasternak:
Panel is default right after import, after close&repoen of the project, the last setting is loaded

PR #558, opened 2020-08-19
Do not call get project.getBasePath

tpasternak:
This method may return incorrect value. The case is that the user may
create project literally *anywhere* and tnen import BSP project to
it. `getBasePath` will return path to the directory where the project
was created.

Apart from that, since this commit the `sources-jar` dir is also
recognized as project internal dependency (updated
`isProjectInternalDependency` method)

### DIFF
--- a/src/com/twitter/intellij/pants/bsp/JarMappings.java
+++ b/src/com/twitter/intellij/pants/bsp/JarMappings.java
@@ -60,11 +60,13 @@ public class JarMappings {
   private synchronized void ensureUpToDate() {
     try {
       if (!librariesFileIsUpToDate) {
-        VirtualFile file = librariesFile();
-        String content = new String(file.contentsToByteArray());
-        Type mapType = new TypeToken<Map<String, String>>() {}.getType();
-        libraryJarToLibrarySourceJar = new Gson().fromJson(content, mapType);
-        librariesFileIsUpToDate = true;
+        Optional<VirtualFile> file = librariesFile();
+        if(file.isPresent()) {
+          String content = new String(file.get().contentsToByteArray());
+          Type mapType = new TypeToken<Map<String, String>>() {}.getType();
+          libraryJarToLibrarySourceJar = new Gson().fromJson(content, mapType);
+          librariesFileIsUpToDate = true;
+        }
       }
     }
     catch (Exception e) {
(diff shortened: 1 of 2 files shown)


tpasternak:
@lukaszwawrzyk

src/com/twitter/intellij/pants/bsp/JarMappings.java
```diff
@@ -60,11 +60,14 @@ public void after(@NotNull List<? extends VFileEvent> events) {
...
-        VirtualFile file = librariesFile();
-        String content = new String(file.contentsToByteArray());
-        Type mapType = new TypeToken<Map<String, String>>() {}.getType();
-        libraryJarToLibrarySourceJar = new Gson().fromJson(content, mapType);
-        librariesFileIsUpToDate = true;
+        Optional<VirtualFile> file = librariesFile();
+        if(file.isPresent()) {
+          String content = new String(file.get().contentsToByteArray());
+          Type mapType = new TypeToken<Map<String, String>>() {
+          }.getType();
```
lukaszwawrzyk, line 67: Maybe fix this awkward formatting ;p

src/com/twitter/intellij/pants/bsp/JarMappings.java
```diff
@@ -125,17 +128,27 @@ private String targetAddressFromSanitizedFileName(String jarName) {
 
   private boolean isProjectInternalDependency(VirtualFile jar) {
     Path jarPath = Paths.get(jar.getPath());
-    Path bloopJarsPath = bloopJarsPath(project);
-    return jarPath.startsWith(bloopJarsPath);
+    Optional<Path> bloopJarsPath = bloopJarsPath(project);
+    Optional<Path> sourcesJarPath = sourcesJarPath(project);
+    return jarPath.startsWith(bloopJarsPath.get()) || jarPath.startsWith(sourcesJarPath.get());
```
lukaszwawrzyk, line 133: are these `get` calls safe?

src/com/twitter/intellij/pants/bsp/JarMappings.java
```diff
@@ -125,17 +128,27 @@ private String targetAddressFromSanitizedFileName(String jarName) {
...
+  private static Optional<Path> bloopJarsPath(Project project) {
+    return PantsBspData.importsFor(project).stream().findFirst()
+      .map(path -> Paths.get(path.getBspPath().toString(), ".bloop", "bloop-jars"));
   }
 
-  private VirtualFile librariesFile() {
-    Path path = Paths.get(project.getBasePath(), ".pants", "libraries.json");
-    return LocalFileSystem.getInstance().findFileByIoFile(path.toFile());
+  private static Optional<Path> sourcesJarPath(Project project) {
+    return PantsBspData.importsFor(project).stream().findFirst()
```
lukaszwawrzyk, line 142: we could have a method to return bloop directory and then this and bloopJarsPath could just resolve single child

PR #562, opened 2020-08-28
Target spec view fixes

tpasternak:
### Commit 1: 
Support non-canonical Pants addresses
Before, the only supported formats were:
  PATH:
  PATH::
  PATH:TARGET

Now we support also address without colon:
  a/b/c
which is equivalent to a/b/c:c

### Commit 2:
Handle the case when LinkedProjectPath is relative
Relative paths were set in case if user has extended project with the "+" button in "Pants"
window

### Commit 3:
Unify failures in findPantsExecutable
As it's return type is Optional, one might not expect it throws

[EDIT][WARNING]:
Commit 6503d33 disables cache. It seems that our cache is too big, so Travis constantly fails to start.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,8 +6,6 @@ language: python
 cache:
   directories:
     - ${HOME}/.cache/pants
-    - .cache/bootstrap # See pants.ini [cache.bootstrap]
-    - .cache/pants     # Pants downloaded by scripts/setup-ci-environment.sh
   timeout: 1200
 notifications:
   email:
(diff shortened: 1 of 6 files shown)


tpasternak:
@wisechengyi would it be possible to merge it? These changes fix some annoying exceptions that are thrown in edge cases

olafurpg:
The CI is green now 🚀

tpasternak:
> The CI is green now 🚀

Unfortunately it's because I disabled cache. Seems that our cache is too big for Travis https://travis-ci.community/t/builds-timeout-during-or-shortly-after-creating-directory-home-travis-cache-sccache/5001/2 . All other PRs are red, too

PR #565, opened 2020-09-07
Make "$PantsFilePathRelativeToBuildRoot$" macro working for sources jar

tpasternak:
For dependencies that are not directly specified in config spec, fastpass
generates "*-sources.jar" files. PantsFilePathRelativeToBuildRoot was
not working for sources in these files.

### DIFF
--- a/src/com/twitter/intellij/pants/macro/FilePathRelativeToBuiltRootMacro.java
+++ b/src/com/twitter/intellij/pants/macro/FilePathRelativeToBuiltRootMacro.java
@@ -6,37 +6,80 @@ package com.twitter.intellij.pants.macro;
 import com.intellij.ide.macro.Macro;
 import com.intellij.openapi.actionSystem.CommonDataKeys;
 import com.intellij.openapi.actionSystem.DataContext;
+import com.intellij.openapi.project.Project;
 import com.intellij.openapi.util.io.FileUtil;
 import com.intellij.openapi.vfs.VfsUtil;
 import com.intellij.openapi.vfs.VirtualFile;
+import com.twitter.intellij.pants.bsp.JarMappings;
+import com.twitter.intellij.pants.bsp.PantsBspData;
+import com.twitter.intellij.pants.bsp.PantsTargetAddress;
 import com.twitter.intellij.pants.util.PantsUtil;
+import org.jetbrains.annotations.NotNull;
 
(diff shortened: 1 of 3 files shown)


src/com/twitter/intellij/pants/macro/FilePathRelativeToBuiltRootMacro.java
```diff
@@ -6,37 +6,80 @@
...
   @Override
   public String getDescription() {
     return "Relative path from build root";
   }
 
+  private Optional<String> jarFilePath(Project project, VirtualFile vFile) {
+    Optional<VirtualFile> jarFile = JarMappings.getParentJar(vFile);
+    if(jarFile.isPresent()) {
+      JarMappings mappings = JarMappings.getInstance(project);
+      Optional<Path> target = mappings
```
olafurpg, line 43: Do we guess the root directory from the name of the jar file? I think it can be more robust if fastpass writes the root directory in the jar file, around here 

https://github.com/scalameta/fastpass/blob/df9359fa7756c6e7f0f062af56fe617201897876/fastpass/src/main/scala/scala/meta/internal/fastpass/pantsbuild/SourcesJarBuilder.scala#L24

If that info doesn't exist in the jar file then we can fallback to this heuristic.

tpasternak: Ok, no it tries to get this info from `*jar!/META-INF/fastpass/source-root`

olafurpg (approved):
LGTM 👍

src/com/twitter/intellij/pants/macro/FilePathRelativeToBuiltRootMacro.java
```diff
@@ -6,37 +6,103 @@
...
+          return Optional.empty();
+        }
+      }
+    }
+    return Optional.empty();
+  }
+
+  private Optional<String> jarFilePathBasedOnMetaInf(VirtualFile vFile) {
+    Optional<VirtualFile> jarFile = JarMappings.getParentJar(vFile);
+    if(jarFile.isPresent()) {
```
lukaszwawrzyk, line 67: you could use `ifPresent` (similar to scala `foreach`)

lukaszwawrzyk (approved):

PR #566, opened 2020-09-09
Don't refresh VFS each time when calling findPantsExecutable

tpasternak:


tpasternak:
@lukaszwawrzyk could you look at this?

lukaszwawrzyk (approved):

PR #571, opened 2020-09-18
Dealias content root directories.

olafurpg:
The goal of this change is to reduce the number of IntelliJ UI freezes
by reducing the number of external file events in IntelliJ.

Previously, content roots for generated sources under
`.pants.d/gen/scrooge` pointed to symbolic links. This could cause
IntelliJ to observe a large number of external file events when users
ran Pants commands from the terminal after a git pull. Now, we dealias
the content roots so that they are not affected by Pants commands that
are executed outside of IntelliJ.


olafurpg:
I was able to reproduce a 6min freeze after installing this change so it doesn't seem to be working as I expected, sadly.

PR #572, opened 2020-09-21
Synchronous refresh

tpasternak:
The `idea dump-shared-index` command does not work when refresh is asynchronous. Indexing starts
when refresh is not finished yet, so no modules are loaded to the project yet. Therefore we need
a way to control it.


lukaszwawrzyk (approved):

tpasternak:
@lukaszwawrzyk I changed trigger value from `1` to `true`

lukaszwawrzyk (approved):

PR #586, opened 2020-10-21
Change import provider ID and increase time window for recommendations

tpasternak:


lukaszwawrzyk (approved):

src/com/twitter/intellij/pants/service/project/FastpassRecommendationsCounter.java
```diff
@@ -18,7 +18,7 @@ public boolean tick(Duration buildLength, LocalDateTime time) {
       }
       Duration sinceLast = Duration.between(lastTick, time);
       lastTick = time;
-      if(sinceLast.toMinutes() < 10 &&
+      if(sinceLast.toMinutes() <= 60 &&
```
olafurpg, line 21: Should we make this number configurable via system properties? That way we won't need to republish the plugin binary in order to tweak the settings.

olafurpg (approved):

PR #590, opened 2020-11-26
Use Python from .venv if present

odisseus:
This pull request is related to the issue #591.

### DIFF
--- a/src/com/twitter/intellij/pants/service/project/PantsResolver.java
+++ b/src/com/twitter/intellij/pants/service/project/PantsResolver.java
@@ -120,7 +120,7 @@ public class PantsResolver {
     if(!python.isPresent()){
       python = Optional.of(new PythonVenvBuilder(myExecutor.getBuildRoot().toString())).map(builder -> {
         String target = mainTargetName(selectedTargets);
-        Path venvDir = Paths.get(projectData.getIdeProjectFileDirectoryPath());
+        Path venvDir = Paths.get(projectData.getIdeProjectFileDirectoryPath(), "venv");
         return builder.build(target, venvDir);
       });
     }
--- a/src/com/twitter/intellij/pants/service/project/PantsResolver.java
+++ b/src/com/twitter/intellij/pants/service/project/PantsResolver.java
@@ -19,6 +19,8 @@ import com.twitter.intellij.pants.PantsBundle;
 import com.twitter.intellij.pants.PantsException;
 import com.twitter.intellij.pants.model.SimpleExportResult;
 import com.twitter.intellij.pants.service.PantsCompileOptionsExecutor;
+import com.twitter.intellij.pants.service.project.model.PythonInterpreterInfo;
+import com.twitter.intellij.pants.service.project.model.PythonSetup;
 import com.twitter.intellij.pants.service.project.model.graph.BuildGraph;
 import com.twitter.intellij.pants.service.project.model.ProjectInfo;
 import com.twitter.intellij.pants.util.PantsConstants;
@@ -28,7 +30,10 @@ import org.jetbrains.annotations.Nullable;
 import org.jetbrains.annotations.TestOnly;
 
 import java.io.IOException;
+import java.nio.file.Paths;
+import java.util.ArrayList;
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 
@@ -80,6 +85,11 @@ public class PantsResolver {
     }
   }
 
+  private boolean hasOnlyDefaultPython(PythonSetup py) {
+    List<PythonInterpreterInfo> environments = new ArrayList<>(py.getInterpreters().values());
+    return environments.size() == 1 && environments.get(0).equals(py.getDefaultInterpreterInfo());
+  }
+
   public void resolve(
     @NotNull Consumer<String> statusConsumer,
     @Nullable ProcessAdapter processAdapter
@@ -87,6 +97,16 @@ public class PantsResolver {
     try {
       String pantsExportResult = myExecutor.loadProjectStructure(statusConsumer, processAdapter);
       parse(pantsExportResult);
+
+      PythonSetup pythonSetup = myProjectInfo.getPythonSetup();
+      if(pythonSetup != null && hasOnlyDefaultPython(pythonSetup)) {
+        PythonVenvFinder finder = new PythonVenvFinder(Paths.get(myExecutor.getProjectDir()).getParent());
+        finder.getEnvironment().ifPresent(py -> {
+          String environmentName = String.format("Python virtual environment (%s)", finder.getVersion().orElse("unknown version"));
+          pythonSetup.getInterpreters().put(environmentName, py);
+          pythonSetup.setDefaultInterpreter(environmentName);
+        });
+      }
     }
     catch (ExecutionException | IOException e) {
       throw new ExternalSystemException(e);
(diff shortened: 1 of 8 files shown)


lukaszwawrzyk (approved):

lukaszwawrzyk (approved):

src/com/twitter/intellij/pants/service/project/PythonVenvBuilder.java
```diff
@@ -0,0 +1,58 @@
...
+
+  private PythonVenvBuilder(Path executable) {
+    this.executable = executable;
+  }
+
+  public PythonInterpreterInfo build(String target, Path venvDir){
+    LOG.info(String.format("Invoking the .venv builder with target %s at %s", target, venvDir));
+    try {
+      Process process = new ProcessBuilder(executable.toString(),
+                                           "--target", target,
```
lukaszwawrzyk, line 33: don't we want to/can't we pass all the target specs specified during the import rather than one? Or we are doing it multiple times for one project (once. per target)?

src/com/twitter/intellij/pants/service/project/PythonVenvBuilder.java
```diff
@@ -0,0 +1,58 @@
...
+      result.setBinary(venvDir.resolve("bin/python").toString());
+      result.setChroot(venvDir.toString());
+      return result;
+    } catch(IOException | InterruptedException ie) {
+      throw new RuntimeException("An error occurred while running the .venv builder", ie);
+    }
+  }
+
+  public static Optional<PythonVenvBuilder> find() {
+    Path executable = Paths.get(System.getProperty("user.home"), "workspace", "source", "dist", executableName); //FIXME
```
lukaszwawrzyk, line 51: what about this fixme? what is the location of venv builder?

src/com/twitter/intellij/pants/service/project/PythonVenvFinder.java
```diff
@@ -0,0 +1,69 @@
...
+
+import com.intellij.openapi.diagnostic.Logger;
+import com.twitter.intellij.pants.service.project.model.PythonInterpreterInfo;
+
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.Optional;
+
+public class PythonVenvFinder {
```
lukaszwawrzyk, line 14: Will the finder find what was generated by the plugin? Is the fined needed at all if we know where we put the env, and which env we need to use?

src/com/twitter/intellij/pants/service/project/resolver/PantsCreateModulesExtension.java
```diff
@@ -112,10 +108,15 @@ public void resolve(
 
     final DataNode<ModuleData> moduleDataNode = projectInfoDataNode.createChild(ProjectKeys.MODULE, moduleData);
 
-    DataNode<ProjectSdkData> sdk = ExternalSystemApiUtil.find(projectInfoDataNode, ProjectSdkData.KEY);
-    if(sdk != null){
-      ModuleSdkData moduleSdk = new ModuleSdkData(sdk.getData().getSdkName());
-      moduleDataNode.createChild(ModuleSdkData.KEY, moduleSdk);
+    if(targetInfo.isPythonTarget()) {
+      // FIXME this happens to work becuse a SDK with the matching name is created later. See PantsPythonSetupDataService.java:77
+      moduleDataNode.createChild(ModuleSdkData.KEY, new ModuleSdkData("python")); //FIXME
```
lukaszwawrzyk, line 113: Is it a good name for sdk if we have separate sdk for project or even some modules? Can we generate sdk with custom name and just set it?

lukaszwawrzyk (changes requested):

src/com/twitter/intellij/pants/service/project/PythonVenvFinder.java
odisseus: The venv_builder tool is invoked only if the finder did not find an existing virtual environment. After a new environment has been created, there is no point in running the finder again.

However, if there's a previously generated environment that was used for VS Code, the finder will detect it, and the venv_builder will not be invoked.

src/com/twitter/intellij/pants/service/project/PantsResolver.java
```diff
@@ -93,6 +104,39 @@ public void resolve(
...
+      });
+    }
+    python.ifPresent(py -> {
+      String environmentName = String.format("Python virtual environment (%s)", finder.getVersion().orElse("unknown version"));
+      pythonSetup.getInterpreters().put(environmentName, py);
+      pythonSetup.setDefaultInterpreter(environmentName);
+    });
+  }
+
+  private String mainTargetName(ProjectData projectData){
```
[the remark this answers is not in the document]
odisseus: Yes, unless there is no target with such name, in which case this method would throw.

src/com/twitter/intellij/pants/service/project/PythonVenvBuilder.java
tpasternak: I think this paths should be found with `findBuildRoot`

src/com/twitter/intellij/pants/service/project/PythonVenvBuilder.java
```diff
@@ -0,0 +1,58 @@
...
+
+  public PythonInterpreterInfo build(String target, Path venvDir){
+    LOG.info(String.format("Invoking the .venv builder with target %s at %s", target, venvDir));
+    try {
+      Process process = new ProcessBuilder(executable.toString(),
+                                           "--target", target,
+                                           "--venv-dir", venvDir.toString(),
+                                           "--pip"
+      ).start();
+      process.waitFor();
```
tpasternak, line 37: Could we add timeout here in case of hang?

src/com/twitter/intellij/pants/service/project/PythonVenvBuilder.java
```diff
@@ -0,0 +1,58 @@
...
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.Optional;
+
+public class PythonVenvBuilder {
+
+  private final Path executable;
+
+  private static final String executableName = "venv_builder.pex";
```
tpasternak, line 21: I'm wondering if we could run here just `./pants run ...` (this would build&run the exec)

src/com/twitter/intellij/pants/service/project/PythonVenvFinder.java
```diff
@@ -0,0 +1,69 @@
...
+      Path config = venv.resolve(configName);
+      Path binary = venv.resolve(binaryName);
+      if(Files.isRegularFile(config) && Files.isRegularFile(binary)) {
+        try {
+          version = Files.lines(config)
+            .filter(l -> l.startsWith(versionPrefix))
+            .map(l -> l.substring(versionPrefix.length()))
+            .findFirst();
+        } catch (IOException e) {
+          LOG.warn(String.format("An error occurred while reading %s", config), e);
```
tpasternak, line 46: Shouldn't this be Log.error?

src/com/twitter/intellij/pants/service/project/PythonVenvBuilder.java
odisseus: The venv_builder tool seems to support only one target argument, but I have found a way around this.

src/com/twitter/intellij/pants/service/project/PythonVenvBuilder.java
odisseus: Done

src/com/twitter/intellij/pants/service/project/PythonVenvFinder.java
odisseus: Ok

odisseus:
This pull request has been superseded by #594.

PR #594, opened 2020-12-24
Use the venv_builder tool to prepare a Python virtual environment

odisseus:


src/com/twitter/intellij/pants/service/project/PantsResolver.java
```diff
@@ -93,6 +100,30 @@ public void resolve(
...
+        PythonInterpreterInfo python = builder.build(target, venvDir);
+        String environmentName = "Python virtual environment from venv_builder";
+        pythonSetup.getInterpreters().put(environmentName, python);
+        pythonSetup.setDefaultInterpreter(environmentName);
+      });
+    }
+  }
+
+  private String mainTargetName(List<String> selectedTargets){
+    //FIXME can there be more than one target?
```
tpasternak, line 123: +1

odisseus: The venv_builder tool supports only one target selector (which, however, might be as general as `::`), so I think this way of selecting the "main" target is an acceptable compromise for now.

lukaszwawrzyk (approved):
Approved for merge. There are minor thinks to fix, primarily related to clean code rather than functioning, that can be fixed. after @odisseus is back from holiday.

common/com/twitter/intellij/pants/util/PantsUtil.java
```diff
@@ -421,6 +422,54 @@ public static GeneralCommandLine defaultCommandLine(@NotNull File pantsExecutabl
     }
   }
 
+  public static Collection<String> listMatchingTargets(@NotNull String projectPath, String selector) throws PantsException {
```
tpasternak, line 425: I think we could use `com.twitter.intellij.pants.bsp.PantsTargetAddress` class here instead of `String` for selector 

(and it could be moved from bsp package)

src/com/twitter/intellij/pants/service/project/PythonVenvBuilder.java
```diff
@@ -0,0 +1,77 @@
...
+  private final String projectPath;
+  private final ProcessAdapter processAdapter;
+
+  private PythonVenvBuilder(String projectPath, ProcessAdapter processAdapter) {
+    this.projectPath = projectPath;
+    this.processAdapter = processAdapter;
+  }
+
+  public static Optional<PythonVenvBuilder> forProjectPath(String projectPath, ProcessAdapter processAdapter){
+    Collection<String> allTargets = PantsUtil.listMatchingTargets(projectPath, "entsec/venv_builder:venv_builder");
```
tpasternak, line 32: Shouldn't the target name be configurable?

tpasternak (approved):
In general I think it's OK. I made some comments, the most important one is about the hardcoded paths

src/com/twitter/intellij/pants/service/project/PythonVenvBuilder.java
odisseus: The target name is always the same. If your project doesn't have this exact target, you probably can't use the venv_builder tool anyway.  Therefore, I think it is permissible to hardcode this target name.

However, the tool itself doesn't seem to contain any special logic. In the future, we might want to get that tool released to the public, or perhaps create a similar one.

common/com/twitter/intellij/pants/util/PantsUtil.java
odisseus: What is the conceptual difference between `com.twitter.intellij.pants.bsp.PantsTargetAddress` and `com.twitter.intellij.pants.model.PantsTargetAddress`?

PR #619, opened 2021-06-16
Enabling external-api.jar generation for IJ 2021.1.1

sellophane:
In order to continue with the update, we must generate new IJ API jar and publish it, so that the build could use it.


lukaszwawrzyk (approved):

wisechengyi (approved):
Thanks, @sellophane !

PR #625, opened 2021-07-29
Fixed git sha

tpasternak:


.travis.yml
```diff
@@ -35,21 +31,15 @@ env:
...
+      env: CACHE_NAME=C2.2 PANTS_SHA="1.25.x-twtr" IJ_ULTIMATE=false PANTS_TEST_JUNIT_TEST_SHARD=1/2 PANTS_PLUGIN_DEPLOY=true
 
 script:
   # Test a single Java target without scala plugin
   - ENABLE_SCALA_PLUGIN=false ./scripts/run-tests-ci.sh --test-junit-test=com.twitter.intellij.pants.integration.OSSPantsJavaExamplesIntegrationTest
   - travis_wait 105 ./scripts/run-tests-ci.sh
 
 after_success:
-  - scripts/deploy/deploy.sh
+  - if [ "PANTS_PLUGIN_DEPLOY" = "true" -a "$TRAVIS_BRANCH" = "master" ]; then scripts/deploy/deploy.sh; fi
```
tpasternak, line 45: I have added this `if` statement here, but not sure if it's the right thing. Please correct me if I'm wrong, but it looks like we have had four jobs running in parallel, and the one that was the fastest was uploading artifacts to the JetBrains repo. @tpasternak @wisechengyi @odisseus

wisechengyi: shouldn't be necessary. For PR builds, there's no creds available for upload.

wisechengyi: and yes for master builds, the fastest shard will upload, but that hasn't been an issue.

wisechengyi: the deploy script already has this logic:
```$ cat scripts/deploy/deploy.sh
#!/bin/bash
# Trigger deploy process if travis ci build is on master or on a tag.
if [ "$TRAVIS_BRANCH" == "master" ] || [ ! -z "$TRAVIS_TAG" ]; then
```

tpasternak:
@wisechengyi Can we push it in this form? Do we still need to test against pants 1.16 and 1.18?

wisechengyi (approved):

PR #627, opened 2021-08-30
Refactor building external api jar; bump to 2021.2

lukaszwawrzyk:


build-external-api-jar/builder/src/main/resources/TestApiJarBuilder.conf
```diff
@@ -1,17 +1,14 @@
 versions {
   intellij {
-    build = "211.7142.45"
-    release = "2021.1.1"
+    build = "212.4416.7"
+    release = "2021.2"
```
tpasternak, line 4: 2021.2.1 has already been released, can we use it?

tpasternak (approved):

build-external-api-jar/builder/src/main/resources/TestApiJarBuilder.conf
lukaszwawrzyk: done

wisechengyi (approved):

PR #635, opened 2021-09-07
Fix Pants build

wisechengyi:


wisechengyi:
Going to merge since it got past the failure. Failed due to later test timeouts.

PR #639, opened 2021-09-08
Migrate build system to Gradle

kamildoleglo:
For publishing to work we have to decide on the naming of environment variables for plugin signing and uploading, and probably create a GHAction, although it should be similar to the test one, just with different Gradle task and env variables. 
A test run for this PR can be viewed on my fork.

Despite some file moves showing up pre-commit as moves correctly, after a commit my git treats them as deletion&addition, regardless of the similarity index value. Not sure if and how I can fix that


wisechengyi:
It appears that it's also removing pants build in the same diff. Do you think it'd be possible to break out the changes into 
1. add gradle build
2. remove pants build

so the changes can be more incremental as well as easier to be reviewed.

Mostly a suggestion. If you think it is difficult to do, please kindly ignore.

.github/workflows/run-tests.yml
```diff
@@ -0,0 +1,42 @@
...
+  build:
+    runs-on: ubuntu-20.04
+    steps:
+      - uses: actions/checkout@v1
+      - name: Prepare
+        run: sudo apt-get install -y openjdk-8-jdk openjdk-11-jdk unzip curl python3-dev python3-virtualenv pkg-config git g++ libssl-dev make g++ zip
+      - uses: actions/setup-java@v2
+        with:
+          distribution: 'zulu'
+          java-version: '11'
```
tpasternak, line 21: We install jdk 11 twice

externalTestApi/build.gradle.kts
```diff
@@ -0,0 +1,35 @@
...
+}
+
+val externalSystemTestApi: Configuration by configurations.creating
+
+fun externalSystemTestApiJar() = zipTree(externalSystemTestApi.singleFile).matching {
+    include("external-system-test-api.jar")
+}
+
+dependencies {
+    externalSystemTestApi("pantsbuild:intellij-pants-plugin:external-system-test-api-v211.7142.45:external-system-test-api@zip")
```
tpasternak, line 31: Version can be parametrized

intellij-pants-plugin.iml
```diff
@@ -1,106 +1,4 @@
 <?xml version="1.0" encoding="UTF-8"?>
```
tpasternak, line 1: I think this file can be removed, as Gradle perfectly replaces DevKit

tpasternak (commented):
Thank you Kamil!!

BTW I think we may remove all files in the .idea directory now

intellij-pants-plugin.iml
tpasternak: And in the follow-up PR we could save "gradle:runIde" runConfiguration in the git

https://www.jetbrains.com/help/idea/run-debug-configuration.html#save-as-permanent

.idea/jarRepositories.xml
```diff
@@ -0,0 +1,70 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
lukaszwawrzyk, line 1: Do we still need to commit .gradle directory, or can IntelliJ just generate it from gradle?

tpasternak:
Unfortunately,`.github/workflows/build-test-jar.yml` is still needed, shouldn't be removed

.github/workflows/run-tests.yml
```diff
@@ -0,0 +1,42 @@
...
+          export DUMMY_REPO_HOME="$CWD/.cache/dummy_repo"
+          export JDK_LIBS_HOME="$CWD/.cache/jdk-libs"
+          ./gradlew :test
+      - name: Save build report if tests failed
+        if: ${{ failure() }}
+        uses: actions/upload-artifact@v2
+        with:
+          name: build-reports
+          path: 'build/reports/'
+          retention-days: 2
```
tpasternak, line 42: Personally, I'd prefer onger retention time and unconditional save

.github/workflows/run-tests.yml
```diff
@@ -0,0 +1,42 @@
+name: Run Gradle tests
+
+on: push
+
+env:
+  artifact_path: /tmp/external-system-test-api.zip
+  PANTS_SHA: "1.25.x-twtr"
```
tpasternak, line 7: Personally I'd prefer to have git SHA instead of branch head here

tpasternak:
Don't forget about releases :)

.github/workflows/run-tests.yml
tpasternak: Also wondering if it would be nice to upload plugin bundle as well

lukaszwawrzyk (approved):

.github/workflows/run-tests.yml
kamildoleglo: Fixed

externalTestApi/build.gradle.kts
kamildoleglo: Done

intellij-pants-plugin.iml
kamildoleglo: Removed

.idea/jarRepositories.xml
kamildoleglo: Removed

.github/workflows/run-tests.yml
kamildoleglo: Extended retention time, added unconditional save

.github/workflows/run-tests.yml
kamildoleglo: This was taken from `travis.yml`, changed to SHA

kamildoleglo:
> Do you think it'd be possible to break out the changes into
> * add gradle build
> * remove pants build

Did that, hope I haven't broken anything

externalTestApi/build.gradle.kts
kamildoleglo: Actually, should this be merged with the IJ version or do we want a possibility to have different test-api.jar and IJ versions?

wisechengyi:
> > Do you think it'd be possible to break out the changes into
> > 
> > * add gradle build
> > * remove pants build
> 
> Did that, hope I haven't broken anything. Not sure that helped though, the worst thing are file relocations seen as addition/deletion

Yeah that's what I meant. So feel free to ignore my suggestion if it turns out difficult.

externalTestApi/build.gradle.kts
lukaszwawrzyk: I think we can, as we tend to avoid bumping the test jar version if it is not necessary. We may end up with bumping a minior version of ij for some reason and leaving test api jar as is.

.github/workflows/publish.yml
```diff
@@ -0,0 +1,43 @@
...
+      - name: Cache .cache
+        uses: actions/cache@v2
+        with:
+          path: .cache
+          key: pants-${{ env.PANTS_SHA }}
+      - name: Setup CI
+        run: PANTS_SHA=${{ env.PANTS_SHA }} IJ_ULTIMATE=${{ env.IJ_ULTIMATE }} ./scripts/setup-ci-environment.sh
+      - name: Test
+        run: |
+          export CWD=$(pwd)
```
lukaszwawrzyk, line 30: minor (worth doing if there is something more to fix). Extract this to something like ./scripts/ci_run_gradle.sh to avoid repeating this exports

.idea/compiler.xml
```diff
@@ -13,11 +11,17 @@
       <entry name="!?*.kt" />
       <entry name="!?*.clj" />
     </wildcardResourcePatterns>
-    <annotationProcessing>
-      <profile default="true" name="Default" enabled="false">
-        <processorPath useClasspath="true" />
-      </profile>
-    </annotationProcessing>
-    <bytecodeTargetLevel target="1.8" />
+    <bytecodeTargetLevel target="11">
```
lukaszwawrzyk, line 14: I thought you deleted .idea from git, is it necessary for anything?

lukaszwawrzyk (approved):

.github/workflows/publish.yml
kamildoleglo: We should migrate all those scripts to Gradle and give them reasonable defaults, this way we can drop those exports

.idea/compiler.xml
kamildoleglo: Thought so too, I must have lost the commit during a rebase, fixed

kamildoleglo:
I've reset to the old branch, because after the rebase and splitting commits, the tests were broken and I couldn't figure out what was wrong, besides the split didn't make this more reviewable

kamildoleglo:
Ping @wisechengyi

PR #641, opened 2021-09-24
Add publication on every push to master, fix publication version

kamildoleglo:


kamildoleglo:
It would be best to avoid duplicating the `publish.yml` action but I couldn't find a clean way of distinguishing between a push and a tag when having a common action for them. If somebody has any better ideas on how to do it, please let me know

.github/workflows/run-tests.yml
```diff
@@ -1,14 +1,14 @@
 name: Run Gradle tests
 
-on: push
+on: [push, pull_request]
```
lukaszwawrzyk, line 3: Not sure, but it might be that we will get double builds, one for pr and one for push. I think it is usually done like: on any pr or on push to master

gradle.properties
```diff
@@ -4,10 +4,10 @@
 # idea_version=211.6693.111
 # scala_plugin_version=2021.1.21
 
+pluginVersion=1.17.0
```
lukaszwawrzyk, line 7: ideally we would push a tag like v1.18.0 and it would publish with a version according to the tag. Current solution is also ok I guess

.github/workflows/run-tests.yml
kamildoleglo: Fixed

gradle.properties
kamildoleglo: I though about that and I can do it this way, however it will require a consistent tagging scheme and I'm not so sure it's the best solution. The plugin version in the properties should always be up to date for bleeding edge releases anyway. 
If you really want me to rework it though, then sure, I'll do

lukaszwawrzyk: As I said current solution is also ok.
Just to discuss, I am not sure how it worked before, we seem to have consistent tagging and bleeding edge versions. In sbt there is a plugin that also correctly assigns version, I think based on latest release tag (if you are on tag it is a release, else it is a snapshot).

lukaszwawrzyk (approved):

wisechengyi (approved):

PR #651, opened 2021-11-02
Remove use of setOption

lukaszwawrzyk:
Remove usages of deprecated API Module.setOption and Module.getOptionValue which triggers unnecessary project refreshes during import. It also removes some of unused methods.
The PR is based partially on [this commit](https://github.com/ESLGaming/intellij-pants-plugin/commit/75573bcef7eb6f24ba32a9cf7c5730475c81d278) but also doesn't break Imported Files view.


wisechengyi (approved):
Thanks, Lukasz!

PR #653, opened 2021-12-02
Update external api jar to 2021.3

lukaszwawrzyk:


build-external-api-jar/run.sh
```diff
@@ -1,6 +1,7 @@
 #!/bin/bash
 
-SOURCES_DIR="/tmp/idea-sources"
```
lukaszwawrzyk, line 3: This directory doesn't exist as idea is cloned by ide-probe and cleaned up after

wisechengyi (approved):

